PR #15, opened 2025-03-19
cmd/containerbuild: set container build timestamp to the timestamp of the commit

Xe:


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Docker images are now built with the timestamp set to the commit timestamp
```
SuperSandro2000, line 10: Since the FROM images are not pinned, I am not sure if this is such a good idea. If I rebuild the commit in the future I might get a different result which might not even work in the absolute worst case.

PR #19, opened 2025-03-19
cmd/anubis: drastically optimize proof of work

### LINKED ISSUE #12: Profile the proof-of-work.mjs file, see if there's low hanging fruit
If this can be sped up, Anubis is faster for everyone.

### LINKED ISSUE #17: The 2000 ms delay after success
Is the 2-second delay after successful validation really needed?

https://github.com/TecharoHQ/anubis/blob/c81e938f63281dbe885f5d3e0ca842e2719de81c/cmd/anubis/js/main.mjs#L67-L70

I think that other solutions like Cloudflare or Vercel do initiate the redirect immediately as soon as the auto-validations were completed, to minimize the time the real user has to wait before seeing the site. Imo it should just redirect immediately as soon as it knows it can do so... (or after like 10 ms or with requestAnimationFrame: in general the user should then still see "Success!" while the actual page is then loading in their browser, which can also take a couple hundred of milliseconds, depending on the speed of the backend app and the user's connection (and currently that time is added on top of those 2000 ms))

Xe:
Closes #12
Closes #17

This drastically optimizes the proof of work check by removing the stringify call at every iteration. Additionally, this optimizes the checks by running them in parallel for as many threads as the browser has available (according to navigator.hardwareConcurrency).

This also changes the redirect lag to 250 milliseconds instead of 2000 milliseconds in order to be perceptually faster. This is below the reaction time threshold of many people, so this will make the post-check success phase perceptually instant.

Testing on an iPhone 7 Plus has shown that this can clear a difficulty 4 check in 3.4 seconds.

This actually optimizes the check so much it may be a logistical concern for operators.

---

To test this, use the Docker image:

```
ghcr.io/techarohq/anubis:pr-19
```

I have deployed this to [git.xeserv.us](https://git.xeserv.us) for testing.


cmd/anubis/js/main.mjs
```diff
@@ -57,15 +59,16 @@ const imageURL = (mood) => {
   const t0 = Date.now();
   const { hash, nonce } = await process(challenge, difficulty);
   const t1 = Date.now();
+  console.log({ hash, nonce });
 
   title.innerHTML = "Success!";
   status.innerHTML = `Done! Took ${t1 - t0}ms, ${nonce} iterations`;
-  image.src = imageURL("happy");
+  image.src = imageURL(`happy?cacheBuster=${anubisVersion}`);
```
jeda: This is resulting in a `happy?cacheBuster=v1.12.1-6-g74ac354.webp` URL (tested on `git.xeserv.us`), which doesn't work and means the image is broken for the ~second it appears on.

Xe: That would explain it! I thought it just loaded so fast that it didn't have time to finish loading lol

Xe:
I think this may be good enough for now. Surely there are going to be needs for things like limiting parallelism, but if that does come to pass we can cross that bridge when we come to it. I think the UX needs of this patch far outweigh the cost to admins.

PR #21, opened 2025-03-19
Explicitely define image sources in Dockerfile

dennis1248:
Explicitely refering to docker.io will make the build succeed on software such as podman which does not default to docker.io as the standard image source.

~~Using 0.0.0.0 instead of localhost fixes Docker not detecting the port the app is binding to. Reverse proxies such as Traefik are now able to automatically detect the port upon which the app is taking connections.~~


cmd/anubis/main.go
```diff
@@ -161,13 +161,13 @@ func main() {
 
 	mux.HandleFunc("/", s.maybeReverseProxy)
 
-	slog.Info("listening", "url", "http://localhost"+*bind, "difficulty", *challengeDifficulty, "serveRobotsTXT", *robotsTxt, "target", *target, "version", anubis.Version)
+	slog.Info("listening", "url", "http://0.0.0.0"+*bind, "difficulty", *challengeDifficulty, "serveRobotsTXT", *robotsTxt, "target", *target, "version", anubis.Version)
```
Xe: An, the intent of this (and the metrics one) is to make the link clickable so it opens in my browser. I'll have to make sure that darwin points 0.0.0.0 to localhost like Linux does.

dennis1248:
Lemme validate if [ccb69a0](https://github.com/TecharoHQ/anubis/pull/21/commits/ccb69a00e2f5bbede6503355db18e4ede7599ba7) actually fixes the issue. I am getting inconsistent results in my testing.

dennis1248:
Ok, seems https://github.com/TecharoHQ/anubis/pull/21/commits/ccb69a00e2f5bbede6503355db18e4ede7599ba7 indeed does not solve the issue. Docker still fails to detect the app binding to a port, see PORTS below, I'd expect `3000/tcp` to appear in the list;

```
CONTAINER ID   IMAGE                                     COMMAND                  CREATED             STATUS                       PORTS              NAMES
46274193e027   667e67366ab2                              "/app/bin/anubis"        20 minutes ago      Up 20 minutes                                   redlib-libreddit-anubis-1
152aafc82c8d   quay.io/redlib/redlib:latest              "redlib"                 2 days ago          Up 2 days                    8080/tcp           redlib-libreddit-1
```

So for now I have to continue manually defining the port Traefik should connect with.

```
- "traefik.http.services.libreddit-anubis.loadbalancer.server.port=3000"
```

@Xe  Do you want to keep this specific commit or pull it? One could argue that 0.0.0.0 is cleaner since it requires no resolving.

Xe:
I'm kinda loathe to change the empty binds to 127.0.0.1 because the empty binds cause Go to listen on all interfaces (IPv4 and IPv6) but 127.0.0.1 only listens on IPv4.

dennis1248:
That makes sense. Since it will be running behind a proxy of some kind which will be using IPv4. I will drop this commit.

SuperSandro2000 (approved):
I have configured my podman to default to docker.io again but probably not everyone has

Xe:
I probably need to remove the Dockerfile because as of #13 it's not used in CI.

dennis1248:
It is nice to have both a Dockerfile and docker-compose.yml inside of the repo itself, as reference material.

But you can opt for adding this info all to the README instead. I do tend to prefer having the files.

Xe:
I rebased on main and am going to merge when CI passes. Thanks!

Xe (approved):

Xe:
This failure is expected. Need to do something to fix PRs not being able to write to GHCR. Annoying.

SuperSandro2000:
> This failure is expected. Need to do something to fix PRs not being able to write to GHCR. Annoying.

You could use conditions like https://github.com/NixOS/nixpkgs/blob/master/.github/workflows/periodic-merge-6h.yml#L21

PR #27, opened 2025-03-20
Ensure content flows and stays centered in small screens

### LINKED ISSUE #18: Misaligned elements on small screens
Most notable on phones, but also on small desktop windows. Tested on Firefox and Safari. Not sure exactly which part of the CSS causes this at a glance, but it seems the page is too wide, and as such the elements aren't centered and get cut off on the right.

Image

Image

Screenshot

dennis1248:
Fixes #18


dennis1248:
Working on this I noticed some deps such as xess are imported from external first party sources, it would simplify development if these files are instead loaded from the project itself.

I also recommend dropping the minimized versions of files, and instead minimize them at build time.

dennis1248:
The image now also resizes with the page itself. It is not very pretty when it does, but I guess this is preferable over overflowing the image.

dennis1248:
Example worst-case scenario;

Schermafdruk van 2025-03-20 16-32-04

SuperSandro2000:
> 2\. go generate ./... && go run ./cmd/anubis

Regarding that: We should probably not commit the minified and compress files to not let the repo grow to much over time.

cmd/anubis/index.templ
```diff
@@ -165,13 +164,13 @@ templ index() {
 	<div class="centered-div">
 		<img
 			id="image"
-			width="256"
+			style="width:100%;max-width:256px;"
```
dennis1248, line 167: I opted for adding these inline, I can move them to the main CSS if this is preferred.

Xe: This is fine!

Xe (approved):

PR #42, opened 2025-03-20
Fix when hardwareConcurrency is undefined

CUB3D:
When `navigator.hardwareConcurrency` is undefined the `Calculating` page will hang forever, making the page impossible to access. Assume at least one thread exists.

Checklist:


Xe:
Just wondering so I can add it to the integration jungle, but what do you do to create this state?

CUB3D:
One way of causing this is using any version of Safari before 15.4, as per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/hardwareConcurrency)

cmd/anubis/js/proof-of-work.mjs
```diff
@@ -1,6 +1,6 @@
 // https://dev.to/ratmd/simple-proof-of-work-in-javascript-3kgm
 
-export function process(data, difficulty = 5, threads = navigator.hardwareConcurrency) {
+export function process(data, difficulty = 5, threads = (navigator.hardwareConcurrency || 1)) {
```
Xe, line 3: TIL about this JS trick!

Xe (approved):
LGTM, thanks so much again!

PR #54, opened 2025-03-21
cmd/anubis: use actual listener address

AlexanderYastrebov:
Use actual listener address in case user requested to choose a random port by ommiting port value or setting it to zero.

Also allow specifying listener address other than localhost, e.g. [::]:0.

Checklist:

- [ ] Tested this at least manually

### DIFF
--- a/cmd/anubis/main.go
+++ b/cmd/anubis/main.go
@@ -111,19 +111,21 @@ func doHealthCheck() error {
 }
 
 func setupListener(network string, address string) (net.Listener, string) {
+	listener, err := net.Listen(network, address)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


AlexanderYastrebov:
Its not correct.

AlexanderYastrebov:
This is correct e.g. httptest.Server [does](https://cs.opensource.google/go/go/+/refs/tags/go1.24.1:src/net/http/httptest/server.go;l=132) but its not convenient for user as logged url (e.g. for `-bind :5555` it may be `http://[::]:5555`) will not work in the browser apparently due [insecure origin](https://stackoverflow.com/questions/57146558/typeerror-digest-of-undefined-in-development-environment). The right way would be to use `-bind localhost:5555` which is a breaking change.

PR #70, opened 2025-03-21
cmd/anubis: derive signing key from configurable secret

### LINKED ISSUE #32: Persist Anubis ed25519 key to allow for multi-instance operation

AlexanderYastrebov:
Add new flag that provides value for Key Derivation Function.

This enables multiple instances and restarts without invalidating existing cookies.

Fixes #32

Checklist:

### DIFF
--- a/cmd/anubis/main.go
+++ b/cmd/anubis/main.go
@@ -39,6 +39,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"golang.org/x/crypto/hkdf"
 )
 
 var (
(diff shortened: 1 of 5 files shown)


cmd/anubis/main.go
```diff
@@ -709,3 +716,16 @@ func serveMainJSWithBestEncoding(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "text/javascript")
 	http.ServeFileFS(w, r, static, "static/js/main.mjs")
 }
+
+func newPrivateKey(secret string) (ed25519.PrivateKey, error) {
+	rr := rand.Reader
+	if secret != "" {
+		rr = hkdf.New(sha256.New, []byte(secret), nil, []byte("ed25519 signing key"))
+	}
```
AlexanderYastrebov, line 724: Note that secret should be a good random value (e.g. `head -c 32 /dev/urandom | base64`), HKDF is used for "key splitting", here to derive a signing key and maybe later to derive other keys from this (master) secret.

Alternative is to add "passphrase" flag and use password-based KDF (e.g. https://en.wikipedia.org/wiki/PBKDF2) to derive keys from it.

Xe:
This is about to need to be refactored and/or redone, a major refactor is in place.

AlexanderYastrebov:
> This is about to need to be refactored and/or redone, a major refactor is in place.

Sure, for #77 you'd need to decide on type of configuration: a struct, a with*-builder or use functional options...

Xe:
It's using a configuration struct. Once #77 is merged it will be extended to include a keypair with one being generated if it doesn't exist.

Xe:
I think the approach in #97 may be better. Key derivation functions from passwords like this could backfire hard, and ed25519 keys are very small.

AlexanderYastrebov:
I agree. I am exploring the idea from a bit different angle and for https://github.com/AlexanderYastrebov/bot-idp I chose to provide fixed secret in base64 `RawURLEncoding` format like in JWT.

PR #75, opened 2025-03-22
fix: installation instructions and example

ccoors:
I had some issues when following the installation guide. The file `/data/cfg/botPolicy.json` is not used automatically. After debugging I found that it is indeed not the default value for the environment variable, so I clarified the documentation.

Also, in the example the `botPolicy.json` is mounted into the `nginx` container, not the `anubis-nginx`, so that could not have worked anyway.

Checklist:


Xe:
Oops, thanks! Adding this to the integration jungle TODO list.

PR #78, opened 2025-03-22
cmd/anubis: amend generic browser challenge to "Gecko"

Xe:
I'm gonna be honest, this is an extreme galaxy brain strategy and I'm not entirely sure if this will pan out. However I got the idea when reading [a community post][0]

If this works, that would be so much funnier than just using "Mozilla" in the rules. I think that this could greatly backfire though, which is why I'm making a pull request and opening this for feedback from the community.

It would be absolutely hilarious if this works though.

[0]: https://github.com/TecharoHQ/anubis/discussions/68#discussioncomment-12583134

Checklist:

### DIFF
--- a/cmd/anubis/botPolicies.json
+++ b/cmd/anubis/botPolicies.json
@@ -390,7 +390,7 @@
     },
     {
       "name": "generic-browser",
-      "user_agent_regex": "(?i:gecko|applewebkit)",
+      "user_agent_regex": "(?i:gecko)",
       "action": "CHALLENGE"
     }
   ],
--- a/cmd/anubis/botPolicies.json
+++ b/cmd/anubis/botPolicies.json
@@ -390,7 +390,7 @@
     },
     {
       "name": "generic-browser",
-      "user_agent_regex": "Mozilla",
+      "user_agent_regex": "(?i:gecko|applewebkit)",
       "action": "CHALLENGE"
     }
   ],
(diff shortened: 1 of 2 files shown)


makew0rld:
Funnily enough, even Safari has `(KHTML, like Gecko)` in its UA so checking for `AppleWebKit` shouldn't even be necessary. 

https://www.useragents.me/

cmd/anubis/botPolicies.json
```diff
@@ -390,9 +390,9 @@
     },
     {
       "name": "generic-browser",
-      "user_agent_regex": "Mozilla",
+      "user_agent_regex": "(?i:gecko|applewebkit)",
```
Xe:
```suggestion
      "user_agent_regex": "(?i:gecko)",
```

docs/docs/CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 - Fixed and clarified installation instructions
+- Changed default challenge logic for "Gecko" or "AppleWebkit" in the User-Agent string
```
Xe:
```suggestion
- Changed default challenge logic for "Gecko" in the User-Agent string
```

Xe:
This will be reopened when the bots start to omit Mozilla from their User-Agent string.

PR #81, opened 2025-03-22
internal/test: introduce integration tests using Playwright

SyrupThinker:
This change introduces automated integration tests that check the blocking behavior of Anubis for various combinations of browser engines and requests.

It makes use of [Playwright](https://playwright.dev/) to provide scriptable access to a browser engine.

## Implementation

The test suite currently expects an already running Playwright server, and connects to it to perform the actual tests.
It then launches multiple browser engines and sends a variety of requests using different `X-Real-Ips` and `User-Agent` headers to an already running Anubis instance.
Said instance finally forwards the requests back to an HTTP server managed by the test suite.

The checks are currently pretty simple:
- we check for the expected cute jackal image, i.e. the action type
- and a timestamp as the protected resource

### Checks

The checks are currently kept simple, to evaluate, if this is a desirable approach for these tests.
Long term the checks should probably be extended to cover user relevant aspects like error or success messages, but the current implementation is already capable of revealing regressions and unexpected behavior.

### Playwright installation

The installation and creation of browser instances could be handled in the test suite itself, and thus be self-contained.
But there are potential concerns like system compatibility, resource utilization, runner hosting and ease-of-use that might affect the approach taken here.

As such I've opted for the option that is compatible with local testing, and that allows some flexibility in where Playwright runs. This way developers that use unsupported systems (like yours truly) can also run the test suite if needed.

### Anubis spawning

The test suite should really also take care of spawning an instance of Anubis to test against.
This way it can ensure, that an instance is present, and that it is set up as needed for the tests.

Given the plans in #9 to modularize the codebase, I've opted to skip this part for now to avoid the need for spawning processes for this purpose.

## Caveats

It is of course somewhat ironic that this is creating an automated test suite that tests a blocking system for automated systems.
As such we are providing some level of instructions on how to "appear legitimate" to the system, which may be undesirable.

These headless browsers also differ slightly from "actual" ones in their behavior, as such they might not meet a desired level of rigorousness needed for this project.

Checklist:

- [x] Tested this automatically


Xe:
Please rebase your branch on main, it will make the CI failures go away. I'll review on monday!

Xe:
Also seriously, thank you so much! This is gonna make "smoke testing" so much easier. I'm gonna work on a proper integration jungle that will be very closed source (to do deep research on headless browsers), but this will help with at least ensuring that it works as it should in the default config.

PR #83, opened 2025-03-22
cmd/anubis: do not apply bot rules if address check fails

SyrupThinker:
During the implementation of #81 I've noticed, that requests using the Kagibot User-Agent were `ALLOW`ed despite coming from an incorrect IP-range. This seems undesirable.

Given that the current conditionals include a check for the address that can never be triggered, I assume the intend is to actually check the range as well?

Checklist:


Xe (approved):

Xe:
Please rebase your branch on main, it will make the CI failures go away.

PR #85, opened 2025-03-22
Hide directory browsing on the static content

SlyEcho:
Demo: https://anubis.techaro.lol/.within.website/x/cmd/anubis/

image

Currently the embedded static content under the `/.within.website/x/cmd/anubis/` path on any Anubis-enabled site will serve all the content as browsable directories.

While the content itself is also accessible using normal means from the HTML code that loads the JavaScript files and images, for exmple, it is still an industry standard to disallow directory browsing on production websites, just to not give adversaries any possible hints as to what kinds of files are available on the server.

As it is now, the built-in `botPolicies.json` is also revealed in this listing. I'm not sure why it would even be necessary to have this data available on the client side? This PR does not address this issue, so the file will still be available, however the directory listings will not be shown.

PS. I'm not really experienced in Go and this solution was derived from some casual Googling.

Checklist:

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+- Hide the directory listings for Anubis' internal static content
 - Changed `--debug-x-real-ip-default` to `--use-remote-address`, getting the IP address from the request's socket address instead.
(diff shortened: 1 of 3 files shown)


Xe:
galaxy brain idea: put empty index.html files in those folders

SlyEcho:
I think 404 errors should maybe also get a fancy error page?

Xe:
Yeah, that can be done later.

Xe (approved):

PR #86, opened 2025-03-22
Change how to make Anubis work without a reverse proxy

SlyEcho:
This is just something that I had to first hack to make local testing work but I saw that the flag `debug-x-real-ip-default` existed later. But I still think that using the remote address is better than just using a fixed string that is passed on the command line or env.

This is something you can take or leave, it doesn't matter to me that much.

Checklist:

### DIFF
--- a/cmd/anubis/main.go
+++ b/cmd/anubis/main.go
@@ -214,7 +214,7 @@ func main() {
 
 	var h http.Handler
 	h = s
-	h = internal.RemoteXRealIP(*useRemoteAddress, h)
+	h = internal.RemoteXRealIP(*useRemoteAddress, *bindNetwork, h)
 	h = internal.XForwardedForToXRealIP(h)
 
 	srv := http.Server{Handler: h}
--- a/cmd/anubis/main.go
+++ b/cmd/anubis/main.go
@@ -45,7 +45,7 @@ var (
 	slogLevel            = flag.String("slog-level", "INFO", "logging level (see https://pkg.go.dev/log/slog#hdr-Levels)")
 	target               = flag.String("target", "http://localhost:3923", "target to reverse proxy to")
 	healthcheck          = flag.Bool("healthcheck", false, "run a health check against Anubis")
-	debugXRealIPDefault  = flag.String("debug-x-real-ip-default", "", "If set, replace empty X-Real-Ip headers with this value, useful only for debugging Anubis and running it locally")
+	useRemoteAddress     = flag.Bool("use-remote-address", false, "read the client's IP address from the network request, useful for debugging and running Anubis on bare metal")
 )
 
 func keyFromHex(value string) (ed25519.PrivateKey, error) {
@@ -214,7 +214,7 @@ func main() {
 
 	var h http.Handler
 	h = s
-	h = internal.DefaultXRealIP(*debugXRealIPDefault, h)
+	h = internal.RemoteXRealIP(useRemoteAddress, h)
 	h = internal.XForwardedForToXRealIP(h)
 
 	srv := http.Server{Handler: h}
@@ -226,7 +226,7 @@ func main() {
 		"serveRobotsTXT", *robotsTxt,
 		"target", *target,
 		"version", anubis.Version,
-		"debug-x-real-ip-default", *debugXRealIPDefault,
+		"use-remote-address", *useRemoteAddress,
 	)
 
 	go func() {
--- a/cmd/anubis/main.go
+++ b/cmd/anubis/main.go
@@ -214,7 +214,7 @@ func main() {
 
 	var h http.Handler
 	h = s
-	h = internal.RemoteXRealIP(useRemoteAddress, h)
+	h = internal.RemoteXRealIP(*useRemoteAddress, h)
 	h = internal.XForwardedForToXRealIP(h)
 
 	srv := http.Server{Handler: h}
--- a/internal/headers.go
+++ b/internal/headers.go
@@ -2,8 +2,8 @@ package internal
 
 import (
 	"log/slog"
+	"net"
 	"net/http"
-	"strings"
 
 	"github.com/TecharoHQ/anubis"
 	"github.com/sebest/xff"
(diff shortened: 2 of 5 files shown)


Xe:
You're gonna need to remake this, a big refactor just hit. Sorry! Should be the last one for a while.

internal/headers.go
```diff
@@ -3,6 +3,7 @@ package internal
 import (
 	"log/slog"
```
Xe, line 4:
```suggestion
	"log/slog"
	"net"
```

internal/headers.go
```diff
@@ -21,16 +22,16 @@ func UnchangingCache(next http.Handler) http.Handler {
...
+// the setting is enabled by the user.
+func RemoteXRealIP(useRemoteAddress *bool, next http.Handler) http.Handler {
+	if useRemoteAddress == nil || *useRemoteAddress == false {
+		slog.Debug("skipping middleware, useRemoteAddress is empty")
 		return next
 	}
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		r.Header.Set("X-Real-Ip", defaultIP)
+		r.Header.Set("X-Real-Ip", strings.Split(r.RemoteAddr, ":")[0])
```
Xe:
```suggestion
		host, _, err := net.SplitHostPort(r.RemoteAddr)
		if err != nil {
			panic(err) // this should never happen
		}
		r.Header.Set("X-Real-Ip", host)
```

internal/headers.go
```diff
@@ -21,16 +22,16 @@ func UnchangingCache(next http.Handler) http.Handler {
...
 }
 
-// DefaultXRealIP sets the X-Real-Ip header to the given value if and only if
-// it is not an empty string.
-func DefaultXRealIP(defaultIP string, next http.Handler) http.Handler {
-	if defaultIP == "" {
-		slog.Debug("skipping middleware, defaultIP is empty")
+// RemoteXRealIP sets the X-Real-Ip header to the request's real IP if
+// the setting is enabled by the user.
+func RemoteXRealIP(useRemoteAddress *bool, next http.Handler) http.Handler {
```
Xe:
```suggestion
func RemoteXRealIP(useRemoteAddress bool, next http.Handler) http.Handler {
```

cmd/anubis/main.go
```diff
@@ -177,7 +177,7 @@ func main() {
 
 	var h http.Handler
 	h = s
-	h = internal.DefaultXRealIP(*debugXRealIPDefault, h)
+	h = internal.RemoteXRealIP(useRemoteAddress, h)
```
Xe:
```suggestion
	h = internal.RemoteXRealIP(*useRemoteAddress, h)
```

[flag](https://pkg.go.dev/flag) will make this always be non-nil

Xe (approved):
Just some style nits, but otherwise it's perfect!

internal/headers.go
```diff
@@ -21,16 +22,20 @@ func UnchangingCache(next http.Handler) http.Handler {
...
+	if useRemoteAddress == false {
+		slog.Debug("skipping middleware, useRemoteAddress is empty")
 		return next
 	}
 
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		r.Header.Set("X-Real-Ip", defaultIP)
+		host, _, err := net.SplitHostPort(r.RemoteAddr)
+		if err != nil {
+			panic(err) // this should never happen
```
soopyc, line 45: this breaks connecting via unix sockets, because the address is apparently just `@`:

```
http: panic serving @: address @: missing port in address
```

maybe this can be a special case for handling?

<details>
<summary>full trace</summary>
<pre>
http: panic serving @: address @: missing port in address
goroutine 36 [running]:
net/http.(*conn).serve.func1()
        /nix/store/cfjhl0kn7xc65466pha9fkrvigw3g72n-go-1.24.1/share/go/src/net/http/server.go:1947 +0xbe
panic({0x8f65c0?, 0xc000200ae0?})
        /nix/store/cfjhl0kn7xc65466pha9fkrvigw3g72n-go-1.24.1/share/go/src/runtime/panic.go:792 +0x132
github.com/TecharoHQ/anubis/internal.RemoteXRealIP.func1({0xabfc00, 0xc00038e000}, 0xc0002ee3c0)
        /home/cassie/build/anubis/internal/headers.go:36 +0x14b
net/http.HandlerFunc.ServeHTTP(0x7f7d084428a8?, {0xabfc00?, 0xc00038e000?}, 0xc000089008?)
        /nix/store/cfjhl0kn7xc65466pha9fkrvigw3g72n-go-1.24.1/share/go/src/net/http/server.go:2294 +0x29
main.main.XForwardedForToXRealIP.func2({0xabfc00, 0xc00038e000}, 0xc0002ee3c0)
        /home/cassie/build/anubis/internal/headers.go:53 +0x156
net/http.HandlerFunc.ServeHTTP(0x46e739?, {0xabfc00?, 0xc00038e000?}, 0xc000185b70?)
        /nix/store/cfjhl0kn7xc65466pha9fkrvigw3g72n-go-1.24.1/share/go/src/net/http/server.go:2294 +0x29
net/http.serverHandler.ServeHTTP({0xc0002ec3f0?}, {0xabfc00?, 0xc00038e000?}, 0x1?)
        /nix/store/cfjhl0kn7xc65466pha9fkrvigw3g72n-go-1.24.1/share/go/src/net/http/server.go:3301 +0x8e
net/http.(*conn).serve(0xc00022c360, {0xac02f8, 0xc000124390})
        /nix/store/cfjhl0kn7xc65466pha9fkrvigw3g72n-go-1.24.1/share/go/src/net/http/server.go:2102 +0x625
created by net/http.(*Server).Serve in goroutine 1
        /nix/store/cfjhl0kn7xc65466pha9fkrvigw3g72n-go-1.24.1/share/go/src/net/http/server.go:3454 +0x485
</pre>
</details>

SlyEcho: I think that in this case the IP should be 127.0.0.1 anyway, because Unix sockets only work on a local machine anyway?

soopyc: yeah, hardcoding `@` addresses to `127.0.0.1` should be fine. i'm not entirely aware of other exotic behavior so this should be enough.

SlyEcho: I did not use the `@` for the solution, but we already know from other flags that we are listening on Unix sockets, so I came up with a solution using that information. It may be possible that there are some other address formats in the future that will cause crashes but it can be fixed at that time.

Xe (approved):

soopyc (approved):

PR #87, opened 2025-03-22
Show a progress bar for the probability of completing the proof of work challenge

jbellerb:
I found the spinner somewhat frustrating, since even though there's no truly accurate way to represent the time to complete a probabilistic process, the difficulty number doesn't give a very meaningful estimate of how long the user should expect to wait. This patch replaces the spinner with a progress bar based on how likely the user would've already completed the challenge. I think this provides a compromise for giving the idea of how long things _should_ take without actually saying how long things _will_ take.

For styling, I tried to match the overall aesthetic and picked colors xess already uses. I still think of it as somewhat temporary and expect it to be redesigned eventually. The performance impact added by having the worker threads report their status to the progress bar is somewhere between 0.8% and 1.5%. I think the added information makes the page feel faster anyways (also there's some room for further optimization in the hash scripts). Nothing added should change the current level of compatibility with older browsers Anubis already provides.

Thanks for making such a cool program!
 
Checklist:

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -41,6 +41,7 @@ Many other small changes were made, including but not limited to:
 - Linting fixes
 - Make dark mode diff lines readable in the documentation
 - Fix CI based browser smoke test
+- Verification page now shows hash rate and a progress bar for completion probability.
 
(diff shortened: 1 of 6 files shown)


jbellerb:
Rebased, updated commits to follow the code-quality guidelines, and split the benchmark tool into a separate PR.

Xe:
Let's have CI see what breaks!

Xe (approved):

PR #91, opened 2025-03-23
chore: remove built binary and prevent accidental addition again

jaredallard:
Removes `main` from the repo and updates `.gitignore` to help prevent accidents like this again. `.test` comes from go test binaries, which I feel is also worth excluding

Omitted changelog because this doesn't feel like it needs one!

Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 .env
 *.rpm
+
+# Go binaries and test artifacts
+main
+*.test


jaredallard:
Sorry for the PR noise, I didn't notice this in my earlier PR 😢

Xe (approved):
We need to add a CI check for things like this

PR #125, opened 2025-03-26
Dockerfile: changed base image to alpine, switched to non-root user during runtime

godlyObeng:
Changed the base image used to build and run the image from debian to alpine as it is smaller and much more secure. You can see the difference from image attached below.
I also switched the runtime user to "anubis" as running as root is not necessary and not secure.

Checklist:

- [x] Used harbor registry to scan and check the project for vulnerabilities and SBOM.

0.1 tag is the image I built using the original Dockerfile and 0.2 is the new image based on this PR's Dockerfile
image


Dockerfile
```diff
@@ -1,19 +1,27 @@
-FROM docker.io/library/golang:1.24 AS build
+FROM docker.io/library/golang:1.24-alpine AS build
```
SuperSandro2000:
```suggestion
FROM docker.io/library/golang:1.24-alpine3.21 AS build
```

Dockerfile
```diff
@@ -1,19 +1,27 @@
...
+RUN apk add --no-cache git
+
 COPY . .
 RUN --mount=type=cache,target=/root/.cache \
   VERSION=$(git describe --tags --always --dirty) \
-  && go build -o /app/bin/anubis -ldflags="-X github.com/TecharoHQ/anubis.Version=${VERSION}" ./cmd/anubis
+  && go build -trimpath -o /app/bin/anubis -ldflags="-X github.com/TecharoHQ/anubis.Version=${VERSION}" ./cmd/anubis
 
-FROM docker.io/library/debian:bookworm AS runtime
+FROM docker.io/library/alpine:3.19 AS runtime
```
SuperSandro2000:
```suggestion
FROM docker.io/library/alpine:3.21 AS runtime
```

SuperSandro2000 (commented):
3.19 is already a bit old

Xe:
This Dockerfile is not used to build Anubis in prod, it's built by [`cmd/containerbuild`](https://github.com/TecharoHQ/anubis/blob/main/cmd/containerbuild/main.go) and [ko](https://ko.build/). I don't think we need to worry about moving it to Alpine.

godlyObeng:
That may be the case but most people will try and build from the Dockerfile if they want to test things locally or just want to use local images. I believe it is good practice to have a good and secure Dockerfile officially. Unless it is removed and people are directed to using [cmd/containerbuild](https://github.com/TecharoHQ/anubis/blob/main/cmd/containerbuild/main.go) and [ko](https://ko.build/) in the docs.

Xe:
Obsoleted by #148. I don't want to ship a Dockerfile anymore.

PR #133, opened 2025-03-27
chore: general linting and update integration tests to be cross-platform

JasonLovesDoggo:
Updated the integration tests to be cross-platform (tested on Windows 11 & ubuntu) 
Some linting fixes & fieldalignement where it didn't look too bad. 

Didn't add changes to unreleased as all changes are already present 

Checklist:


internal/test/playwright_test.go
```diff
@@ -121,14 +124,15 @@ func run(t *testing.T, command string) string {
...
 	t.Helper()
 
-	shPath, err := exec.LookPath("sh")
-	if err != nil {
-		t.Fatalf("[unexpected] %v", err)
+	var cmd *exec.Cmd
+	if runtime.GOOS == "windows" {
+		cmd = exec.Command("cmd", "/C", "start", "/B", command)
+	} else {
+		cmd = exec.Command("sh", "-c", command+" &")
```
Xe, line 131: This won't do what you think it will I don't think?

JasonLovesDoggo: Hmm, I tested it on windows/linux, and it worked fine. I think my only difference was the background (via & and /B), I can run the previous tests as well to double check they behave the same

JasonLovesDoggo:
Bump! The integration tests not working on windows is pretty annoying

Xe:
I'm not sure I want to merge this as-is. I understand the desire to do this testing from windows, but there's significant semantic differences between cmd.exe and bash syntax in ways that would make it annoying to write. I personally do all my development for Anubis in WSL (specifically in an Ubuntu container). I think it may be better to just tell people to use WSL to do this for now.

I do want to have better windows support eventually, but Windows needs dedicated experience and knowledge in ways that I just don't have the time for. I was personally planning on making windows support first-class when there's an enterprise customer with windows specifically in mind.

Well that, and the fact that Windows server costs enough that people are not willing to run it for open source projects. I'd love to be proven wrong though!

JasonLovesDoggo:
> I'm not sure I want to merge this as-is. I understand the desire to do this testing from windows, but there's significant semantic differences between cmd.exe and bash syntax in ways that would make it annoying to write. I personally do all my development for Anubis in WSL (specifically in an Ubuntu container). I think it may be better to just tell people to use WSL to do this for now.
> 
> I do want to have better windows support eventually, but Windows needs dedicated experience and knowledge in ways that I just don't have the time for. I was personally planning on making windows support first-class when there's an enterprise customer with windows specifically in mind.
> 
> Well that, and the fact that Windows server costs enough that people are not willing to run it for open source projects. I'd love to be proven wrong though!

Fair enough. When that enterprise client comes knocking, feel free to ping me, and I'm happy to work on that. I'll open a PR with just the linting changes sometime tomorrow.  

Thanks for the response!

JasonLovesDoggo:
(though maybe we should update the integration_test.go file to have !goos=windows buildflag

Xe:
Makes sense. Ship it!

PR #136, opened 2025-03-27
experiment: Redirect flow

SlyEcho:
This PR changes the flow a little bit. The JS challenge will not be done at the same URL as the original page, but the user will be redirected to a separate page where the hashing happens. Also, the redirect back to the website happens using a form POST and another redirect, instead of setting the page location from JavaScript.

### The benefits:
 - Could possibly help with CDN's not caching the Anubis HTML content and breaking things.
 - Since everything happens with redirect-post-redirect there will not be any back button weirdness.
 - Separate request to a known URL in the logs to see when the challenges happen.
 - Possible to host Anubis on a different (sub)domain, but getting the cookies working may be not that easy.
 - Needed also to plug into [external authentication for ingress-nginx](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#external-authentication), for example (but there is some more things needed).
 
### Todos:
 - [ ] Gotta check the redirect URL a little better
 - [ ] Should really not pass all that much info through the HTML and URLs at all, but keep state in the server
 - [ ] Probably the POST request to `make-challenge` could be removed as well and the info it returns just inserted to the page some how.

### Checklist:


AlexanderYastrebov:
Hello, check this out https://github.com/AlexanderYastrebov/bot-idp - it implements OpenID Code Flow but more importantly you can pair it with [Skipper](https://github.com/zalando/skipper/) (see [example config](https://github.com/AlexanderYastrebov/bot-idp/blob/main/doc/skipper.yaml)) which provides megatons of features :rocket:

SlyEcho:
> Hello, check this out https://github.com/AlexanderYastrebov/bot-idp - it implements OpenID Code Flow but more importantly you can pair it with [Skipper](https://github.com/zalando/skipper/) (see [example config](https://github.com/AlexanderYastrebov/bot-idp/blob/main/doc/skipper.yaml)) which provides megatons of features 🚀

Yeah, that's very cool stuff and I will definitely have to add that to my things to play with.

I also thought about OIDC, for example the JWT token's key could be made available in a `jwks.json` endpoint and so on.

But I **think** that Anubis is not really an _authentication_ provider, or _identity_ provider. I my **opinion** the cookie should not even be a JWT but some opaque ticket, however that would require some kind of persistence and HA which would make deployment more difficult.

Also, the [forward auth](https://doc.traefik.io/traefik/middlewares/http/forwardauth/)-type schemes in Traefik and nginx are not doing OIDC verification or OAuth2 flow, they are much simpler and would require something like [oauth2-proxy](https://github.com/oauth2-proxy/oauth2-proxy) to be used with a proper auth flow.

AlexanderYastrebov:
> But I think that Anubis is not really an authentication provider, or identity provider. 

Anubis is all-in-one solution that is easy to plug into existing small setups. 

`bot-idp` is for large or corporate systems where it might be easier to integrate via OIDC and avoid whole host of testing and compliance for the reverse proxy part as it will rely on the existing one.

But one may pack Skipper and bot-idc into a container and end up with all-in-one solution.

> I my opinion the cookie should not even be a JWT but some opaque ticket, however that would require some kind of persistence and HA which would make deployment more difficult.

The nice part about bot-idp is that it offloads session handling to the reverse-proxy.
Both Skipper and oauth2-proxy use encrypted cookie by default, Skipper also supports splitting large cookie in chunks.
Other proxies (e.g. oauth2-proxy) may support other session storage, e.g. in Redis.
Re-implementing all of that and a decent reverse proxy on top in Anubis might become quite a challenge.

SlyEcho:
This isn't needed any more and most of the points have been covered by other changes.

PR #137, opened 2025-03-27
Fix broken link to default policy file

Hans5958:
This PR fixes the broken link to https://github.com/TecharoHQ/anubis/blob/main/data/botPolicies.json, possibly after work on #77. Found while reading the docs for random reading.

For the maintainer(s) and contributors, thank you for the work!


Xe:
Thanks for the fix! I thought Docusaurus would warn me about this but apparently not.

Xe (approved):

PR #138, opened 2025-03-27
Added header to redirect to the desired target instead of hardcoded TARGET

MarcUs7i:
Added header to redirect to the desired target instead of hardcoded TARGET (TARGET will be used as default, while the header is the override)
Works perfectly fine on bare metal and docker with network_mode set to "host"

Checklist:

- [X] Tested this at least manually


lib/anubis.go
```diff
@@ -178,6 +206,25 @@ func (s *Server) MaybeReverseProxy(w http.ResponseWriter, r *http.Request) {
 		"x-real-ip", r.Header.Get("X-Real-Ip"),
 	)
 
+	targetBackend := r.Header.Get("X-Target-Backend")
```
AlexanderYastrebov, line 215: We also had this SSRF via X-Skipper-Proxy https://github.com/zalando/skipper/security/advisories/GHSA-f2rj-m42r-6jm2

Xe:
I don't understand the point of this change. What flows does this enable?

lib/anubis.go
Kizuren: Thanks for the info, the issue should be resolved now.

MarcUs7i:
> I don't understand the point of this change. What flows does this enable?

Sorry for not clarifying this beforehand!
Instead of one target `TARGET` it returns to the specified service in the `X-Target-Backend`. I made this change so I can run multiple services through Anubis. If the header is empty, it uses the `TARGET` as usual.

Now, the SSRF vulnerability should be fixed from the use of the `X-Target-Backend` header. It checks now a json file if Anubis should really forward. If the service specified in the header isn't found in the json file, then a error message is displayed.
Btw, I put a watchdog for that file, so there is no need to restart the docker container

AlexanderYastrebov:
> I made this change so I can run multiple services through Anubis.

This is how you achieve this with https://github.com/AlexanderYastrebov/bot-idp (+Skipper) by modifying [example config](https://github.com/AlexanderYastrebov/bot-idp/blob/main/doc/skipper.yaml):
```yaml
# other configs omitted for brevity
inline-routes: |
  // Handle requests to foo.test host
  foo: Host("foo.test")
      // Configure bot-idp as an OpenID Connect Provider,
      -> oauthOidcAllClaims("http://localhost:4159",
          "bot-idp", "secret1",
          "http://skipper.localtest.me:9090/oauth2/callback",
          "openid", "sub")

      // Service1 backend address
      -> "https://service1.test";
      
  // Handle all other requests
  rest: * 
      -> oauthOidcAllClaims("http://localhost:4159",
          "bot-idp", "secret1",
          "http://skipper.localtest.me:9090/oauth2/callback",
          "openid", "sub")

      // Default backend address
      -> "https://service2.test";
```

MarcUs7i:
> This is how you achieve this with https://github.com/AlexanderYastrebov/bot-idp (+Skipper) by modifying [example config](https://github.com/AlexanderYastrebov/bot-idp/blob/main/doc/skipper.yaml)

Thanks? But I am unsure how this is supposed to help for the PR to add similar functionality

Xe:
> I made this change so I can run multiple services through Anubis.

I think that the fix in https://github.com/Xe/x/pull/694 being ported over (after the error handling change) would potentially be better because it would use nginx/traefik subrequest auth.

JasonLovesDoggo:
> he service specified in the header isn't found in the json file, then a error message is displayed.
> Btw, I put a watc

Ideally, this could also be fixed by the caddy/traefik/other proxy/webserver plugins. IMO the scope of Anubis should be to validate users (i.e. libanubis) while the rest is a thin wrapper around it for when you are not using a dedicated webserver.

MarcUs7i:
> I think that the fix in [Xe/x#694](https://github.com/Xe/x/pull/694) being ported over (after the error handling change) would potentially be better because it would use nginx/traefik subrequest auth.

Yes, it would have been better to leave it to nginx/traefik and actually it's cleaner approach. I didn't see that PR. 
I tried to implement the routing directly into anubis, but I agree the fix [Xe/x#694](https://github.com/Xe/x/pull/694) is better.

Xe:
I think this exact change is out of scope pending proper integration with other tools.

PR #148, opened 2025-03-28
all: do not commit generated JS/CSS to source control

### LINKED ISSUE #40: Don't commit generated files?
From [@SuperSandro2000](https://github.com/SuperSandro2000) in #27:

> Regarding that: We should probably not commit the minified and compress files to not let the repo grow to much over time.

Generally it's considered best practice to run `go generate` on your codebase before committing things. Should I make CI run `go generate` before building?

Xe:
Closes #125
Closes #40

Among other things, this moves all of the asset generation to run within the context of an npm script. Developer documentation stubs have been added so that people can get started more easily.

The top-level Dockerfile (which is no longer used in production) has been removed as its presence has been causing confusion. This changeset will break it anyways.

These changes will make for less "repo churn" as the static assets are built and rebuilt, at the cost of making the build step more complicated for downstream packagers. If this becomes a burden, we can explore making a "release tarball" that contains pre-massaged outputs.

Checklist:

- [x] Ran integration tests `npm run test:integration`


kotx:
@Xe I think this breaks dependents on `github.com/TecharoHQ/anubis/lib`? Specifically `/.within.website/x/cmd/anubis/static/js/main.mjs` (or anything under that dir) 404s because it isn't generated when built as a library.

PR #155, opened 2025-03-30
Debug tool for benchmarking proof-of-work algorithms

jbellerb:
While developing #87, I made a benchmarking tool to check that I hadn't introduced any performance regressions. It's fairly basic, only measuring the time a challenge took vs the number of hashes it had to compute. It can also alternate between two algorithms and compare their speeds, but there's high variance between runs so it takes a while to converge on a good number. Huge amount of inline CSS to avoid shipping extra CSS on the non-benchmark pages, otherwise the templates would need a refactor.

Checklist:

- [x] Ran integration tests `npm run test:integration`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+web/index_templ.go linguist-generated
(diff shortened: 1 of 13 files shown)


Xe (approved):
This is really cool. It'll be super useful as I port the challenge to Rust!

jbellerb:
Rebased. Excited for the new challenge!

Issue #131, opened 2025-03-26
Support of Open Graph tags


garfieldairlines:
I'm trying out Anubis and its potential downsides left and right and one thing that I discovered is that Open Graph tags are not working anymore.

A simple workaround would be to add a variable in the docker container for og:image og:url og:description and so on or else link previewers like in Telegram* or Bluesky (and many others) will just fail (as intended).

*(I added them in my IP whitelist afterwards)

JasonLovesDoggo:
> I'm trying out Anubis and its potential downsides left and right and one thing that I discovered is that Open Graph tags are not working anymore.
> 
> A simple workaround would be to add a variable in the docker container for og:image og:url og:description and so on or else link previewers like in Telegram* or Bluesky (and many others) will just fail (as intended).
> 
> *(I added them in my IP whitelist afterwards)

Would you suggest that Anubis initially fetches the raw page directly, parses the of tags out and forwards to the challenge page? If so, how do you suggest caching on that works?

JasonLovesDoggo:
> > > I'm trying out Anubis and its potential downsides left and right and one thing that I discovered is that Open Graph tags are not working anymore.
> > > A simple workaround would be to add a variable in the docker container for og:image og:url og:description and so on or else link previewers like in Telegram* or Bluesky (and many others) will just fail (as intended).
> > > *(I added them in my IP whitelist afterwards)
> > 
> > 
> > Would you suggest that Anubis initially fetches the raw page directly, parses the of tags out and forwards to the challenge page? If so, how do you suggest caching on that works?
> 
> Hi ! Of course not ! But just docker tags that you can edit to show basic information.
> 
> Example : project logo, with "This is the gitlab of $project" and the canonical URL. That's it. But I gotta admit you were thinking further away than me and caching these from the original page sounds interesting… and also technically challenging.

My only issue with "hardcoded" tags (though it's 100000% better than nothing) is that many pages have unique social previews (e.g. this gh repo vs golang/go)

Xe:
This is important enough that we should just fix it by making a decaymap of the downstream tags and serving them in the challenge page. As far as I understand, this would fix the problem completely.

PR #178, opened 2025-03-31
data/botPolicies: allow iMessage scraper by default

Xe:
Allows social preview links on iMessage. This is a HACK until #131 can be realized.

Checklist:


JasonLovesDoggo:
I've got a question, why do we only match Gecko user agents? Logically those are browsers or something pretending to be a browser. But why don't we match something that shamelessly identifies itself as a scraper? 

This PR allows stuff to bypass Anubis but so is just not including gecko in your user agent...?

PR #180, opened 2025-03-31
lib/anubis: actually check the result with the correct difficulty

SlyEcho:
I noticed that the server side is checking the prefix length with this variable `ChallengeDifficulty` that is not set anywhere so it was always 0, meaning it was checking a prefix string of `""` which always passes.

Fixed to use the difficulty number from the detected rule.

### Checklist:


Xe:
This is being tracked as a recurrence of GHSA-56w8-8ppj-2p4f / CVE-2025-24369.

lib/anubis_test.go
```diff
@@ -36,7 +37,7 @@ func spawnAnubis(t *testing.T, opts Options) *Server {
 
 func TestCookieSettings(t *testing.T) {
 	pol := loadPolicies(t, "")
-	pol.DefaultDifficulty = 0
+	pol.DefaultDifficulty = 4
```
Xe, line 61: This probably needs a dedicated test, EG:

```go
func TestCVE2025_24369(t *testing.T)
```

Also please check the email listed on your github profile :)

Xe (approved):

PR #192, opened 2025-04-01
lib: minimize the amount and type of data collected

Xe:
Checklist:

These changes aren't easy to test mechanically, but testing the logging package is probably out of scope.


JasonLovesDoggo:
Are you sure that limiting data collection is the best way forward? PoW is excellent but there will surely be ways to bypass it if Anubis becomes popular enough. Wouldn't it be good to collect additional datapoints in hopes of finding ways to **identify** headless browsers/scrapers?

Xe:
Sadly, this logging is load-bearing. I will retry this later.

PR #199, opened 2025-04-02
Shell: change shebang, bash is not needed

zilti:
Checklist:

### DIFF
--- a/web/build.sh
+++ b/web/build.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 set -euo pipefail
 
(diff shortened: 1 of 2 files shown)


SuperSandro2000:
The problem with that is, that it is undefined what /bin/sh is. On some systems it is bash but on other dash, busybox sh or ksh.

Xe (approved):

Xe:
Bash is needed apparently.

Xe:
TBH Bash is a reasonable part of the dependency closure of modern programs. It's also worth noting that I do my development on a MacBook with macOS ancient bash and this works fine there.

PR #220, opened 2025-04-04
Add 'Opera' to 'generic-browser' bot policy rule

mjeanson:
After deploying Anubis bot traffic is drastically reduced but I still see a lot of requests from User-Agents that claim to be 'Opera' like so:

"Opera/9.90.(Windows NT 6.0; mt-MT) Presto/2.9.173 Version/10.00"
"Opera/8.46.(X11; Linux i686; fo-FO) Presto/2.9.161 Version/11.00"

Add 'Opera' to the generic-browser rule to also challenge them.


Xe:
This is quite interesting. What ISO code is mt and fo?

mjeanson:
They seem to generate random User Agents from real values, the ISO codes, platform and versions all vary for each request.

unixfox:
Do note that Opera 9 and 8 dates back from 2006: https://en.wikipedia.org/wiki/History_of_the_Opera_web_browser#Version_9

So it is highly possible that it won't be compatible with anubis and you might block legitimate users.

Dryusdan:
Hum, from my experience, during a DDoS on my forge I receive many requests (logic) with different user-agent, and Opera (bellow) are in. 

I'm pretty sure I haven't any user with Opera browser on a Windows 95 (or Android 1.1).

I link all unique browser that hit my forge during this DDoS :) 

[bW1.txt](https://github.com/user-attachments/files/19661161/bW1.txt)

Xe:
That is such an odd attack. We may need to add more key words like that yeah.

Xe (approved):

PR #232, opened 2025-04-06
dev: Improvements to build scripts

SlyEcho:
- Rewrite Makefile a little bit to not call into the package.json commands
 - Add esbuild as a package so that it does not need to be installed on the build system
 - Add the node packages to PATH when calling the Bash scripts
 - Update the NPM package lockfile.

Right now this works on a Golang Alpine container:

```sh
# on host:
docker run --rm -it golang:alpine

# on container:
apk add bash brotli git make npm zstd

git clone https://github.com/SlyEcho/anubis.git -b build-commands /src
cd /src

make
./var/anubis
```

Checklist:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,27 +1,30 @@
-NODE_MODULES = node_modules
-VERSION := $(shell cat ./VERSION)
+VERSION= $(shell cat ./VERSION)
+GO?= go
+NPM?= npm
 
 .PHONY: build assets deps lint prebaked-build test
 
-assets:
-	npm run assets
-
-deps:
-	npm ci
-	go mod download
-
-build: deps
-	npm run build
-	@echo "Anubis is now built to ./var/anubis"
-
 all: build
 
-lint:
-	go vet ./...
-	go tool staticcheck ./...
+deps:
+	$(NPM) ci
+	$(GO) mod download
+
+assets: deps
+	$(GO) generate ./...
+	./web/build.sh
+	./xess/build.sh
+
+build: assets
+	$(GO) build -o ./var/anubis ./cmd/anubis
+	@echo "Anubis is now built to ./var/anubis"
+
+lint: assets
+	$(GO) vet ./...
+	$(GO) tool staticcheck ./...
 
 prebaked-build:
-	go build -o ./var/anubis -ldflags "-X 'github.com/TecharoHQ/anubis.Version=$(VERSION)'" ./cmd/anubis
+	$(GO) build -o ./var/anubis -ldflags "-X 'github.com/TecharoHQ/anubis.Version=$(VERSION)'" ./cmd/anubis
 
-test:
-	npm run test
+test: assets
+	$(GO) test ./...
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ deps:
 	$(NPM) ci
 	$(GO) mod download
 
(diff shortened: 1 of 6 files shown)


web/build.sh
```diff
@@ -32,9 +32,9 @@ THE SOFTWARE.
 for the JavaScript code in this page.
 */'
 
-esbuild js/main.mjs --sourcemap --bundle --minify --outfile=static/js/main.mjs "--banner:js=${LICENSE}"
+npx esbuild js/main.mjs --sourcemap --bundle --minify --outfile=static/js/main.mjs "--banner:js=${LICENSE}"
```
Xe:
```suggestion
esbuild js/main.mjs --sourcemap --bundle --minify --outfile=static/js/main.mjs "--banner:js=${LICENSE}"
```

The reason the build scripts are run inside NPM build commands is so that you don't need to use `npx`, which is honestly a massive security problem waiting to happen. This won't work inside the packaging environments that most distros use.

Xe: Maybe the answer is to append the PATH variable in the makefile to include `node_modules/.bin` as the _last_ element in $PATH?

SlyEcho: But they are not run in the NPM build commands, they are run in the bash scripts.

Xe: Yeah but when you run a command from a script via `npm run`, it injects the node_modules bin folder in the PATH:

```json
"env": "env | grep ^PATH="
```

```console
$ npm run env

> @techaro/anubis@1.0.0-see-VERSION-file env
> env | grep ^PATH=

PATH=/home/xe/code/TecharoHQ/anubis/node_modules/.bin:/home/xe/code/TecharoHQ/node_modules/.bin:/home/xe/code/node_modules/.bin:/home/xe/node_modules/.bin:/home/node_modules/.bin:/node_modules/.bin:/home/linuxbrew/.linuxbrew/lib/node_modules/npm/node_modules/@npmcli/run-script/lib/node-gyp-bin:/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin:/home/xe/.local/bin:/home/xe/bin:/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/home/xe/go/bin
```

This means that `postcss` and/or `esbuild` just show up in $PATH and work like you installed them as OS-level dependencies.

SlyEcho: Found another way.

SlyEcho:
Confirmed that the AUR package can also be built, however it could be simplified now because it wouldn't depend on esbuild and postcss now.

Makefile
```diff
@@ -1,27 +1,31 @@
-NODE_MODULES = node_modules
-VERSION := $(shell cat ./VERSION)
+VERSION= $(shell cat ./VERSION)
+GO?= go
+NPM?= npm
```
SuperSandro2000, line 3: I am not sure that this is necessary. Why not rely on the npm/go from PATH?

SlyEcho: It is probably not needed, I was at first testing with a Go binary tarball downloaded to some arbitrary path.

But what it could be helpful for is versioned toolchains and then you could do `GO=go1.24 make`

SuperSandro2000: Side note: you should do ``make GO=go1.24`` as otherwise make falls back to the env which not always works in all circumstances with all makefiles.

Xe (approved):

Issue #94, opened 2025-03-23
Use a better proof-of-work mechanism


Xe:
SHA-256 works for now to buy time but it's not GPU-safe. Use a better mechanism to be sure it's GPU-safe.

runxiyu:
Another problem is that the WebCrypto/SubtleCrypto/IForgotWhatTheProperNamesAre is extremely slow; a simple C implementation outperforms those by 200 times. I found a hash difficulty of 21 bits to be appropriate, but that would take ages in JS

runxiyu:
I think variable delay functions might be worth investigating here.

fossifousacid:
Would switching to a more complicated hash like argon2 help? There are existing implementations in Rust that can be compiled to a webassembly module.

https://www.rfc-editor.org/rfc/rfc9106.html

runxiyu:
Argon2 is too computationally expensive to verify on the server side.

mimi89999:
SHA2 has a WASM SIMD implementation: https://github.com/RustCrypto/hashes/blob/sha2-v0.11.0-pre.5/sha2%2Fsrc%2Fsha256.rs#L30

Any algorithm that doesn't have it will be much worse in the browser than run directly on hardware.

runxiyu:
I just grabbed a generic public domain C implementation of SHA-256. It's significantly slower than running the same program directly (150% vs 200% speedup compared to SubtleCrypto — strange as I'd expect the latter to be optimized too), but still good enough for the purposes of Anubis/etc, in my opinion.

Interesting WASM SIMD implementations.

runxiyu:
See also: https://people.csail.mit.edu/rivest/pubs/RSW96.pdf

nbraud:
> Would switching to a more complicated hash like argon2 help?

Potentially yes.  In general, a memory-hard proof-of-work mechanism would reduce the advantage in utilizing more-specialized hardware, like GPU- or ASIC-based solvers.  That's all-the-more important given that it seems (to me, based on a cursory analysis) attackers could repurpose Bitcoin-mining ASICs (which may be gathering dust somewhere) to attack the current PoW system.

The harder part is making such a PoW system not require too much work on the verifier's side, as it otherwise becomes an easy DoS vector in addition to being a server-side performance issue.

Something like Argon2d might work, but will require as much memory on the verifier side, as one solver thread; intuitively, a PoW protocol which is memory-hard only for the solver might not exist:
[...]

Issue #106, opened 2025-03-24
Difficulty scales poorly


sfan5:
The default of 4 seems to take about 1-2 seconds on my desktop machine.
While I have seen another site use a difficulty of 5 and take 20+ seconds.

Since the leading zeros are checked on the hex-string and not bit-based, increasing difficulty by one will require 4 more zero bits, which increases the computational complexity **16** times.
For more gradual scaling it should check the number of leading zero *bits*.

JasonLovesDoggo:
#240 fixes this

nbraud:
> Wouldn't that still be kinda exponential?

The likelyhood of success, for a random solution, would be `p = 1 - threshold/2ˡ` (with `l` the hash length)
Assuming that successive trials are (statistically) independent, the expected number of attempts before success would be `1/p`.

So it's not exponential, but you are right that it's not quite linear (I was thinking likelyhood, not expectation)
If we want the difficulty to be proportional to the (expected) number of attempts, that would be `difficulty = K/p`, i.e. `threshold = 2ˡ (1 - K/difficulty)`, for some scaling factor `K`.

compilade:
I think it would be possible to use fractional bits as the difficulty, to allow for much more fine-grained steps, even if still exponential. At least it should be relatively intuitive that a difficulty of `15.5` bits is somewhere between `15` bits and `16` bits.

See <https://gist.github.com/compilade/9b16624b81f93fece2637347c27d0bd3> and [the example output (in a comment of that gist)](https://gist.github.com/compilade/9b16624b81f93fece2637347c27d0bd3?permalink_comment_id=5551344#gistcomment-5551344) for a proof of concept of converting a floating-point difficulty to the corresponding max acceptable hash value (in C, but `max_hash_f64` should be easy to adapt to other languages, as long as there is some way to extract the exponent and mantissa of a floating-point number (e.g. by manually unpacking the bits (see <https://float.exposed/>), or with [`frexp(3)`](https://manpages.debian.org/bookworm/manpages-dev/frexp.3.en.html)))

Verifying whether a hash is small enough can then be done relatively simply with the equivalent of [`memcmp(3)`](https://manpages.debian.org/bookworm/manpages-dev/memcmp.3.en.html) between the hash buffer and the max hash from the difficulty.

[...]

PR #240, opened 2025-04-09
Port checker to WASM

Xe:
This is for the pivot strategy to accomplish #94. Overall the strategy looks like this:

1. Port existing checker to WebAssembly
2. Hook it into the existing flow/logic
3. Switch default checker to use this new webassembly-based flow
4. Implement a small set of other checkers (equi-x, etc) and randomly pick between them based on bytes in the challenge.

NOTE: this changes the proof of work calculation logic for the new checker.

This differs from the JavaScript implementations by constructing the hash differently. In JavaScript implementations, the SHA-256 input is the result of appending the nonce as an integer to the hex-formatted challenge, eg:

```js
sha256(`${challenge}${nonce}`);
```

This **does work**, however I think that this can be done a bit better by operating on the challenge bytes _directly_ and treating the nonce as a salt.

The nonce is also randomly encoded in either big or little endian depending on the last byte of the data buffer in an effort to make it more annoying to automate with GPUs.


max-te:
Hi, so I have implemented something like this with Rust and WASM before for proof-of-work on login requests. That code lives in my private repo, but I've contributed some work back to the sha2 crate: If you want to use the SHA2 family of hashes, I recommend you check out the 0.11 pre-release version of the sha2 crate and compile the wasm with `RUSTFLAG=-C target-feature=+simd128`. This activates the [wasm32 simd backend](https://github.com/RustCrypto/hashes/pull/562) which gives around an 8x speed-up..

Xe:
I'm gong to have to see if wazero supports SIMD, if it does then this will be an obvious addition. Thanks!

Xe:
```
go tool benchcmp before-simd.txt after-simd.txt 
benchcmp is deprecated in favor of benchstat: https://pkg.go.dev/golang.org/x/perf/cmd/benchstat
benchmark              old ns/op     new ns/op     delta
BenchmarkSHA256-32     13222850      8209668       -37.91%

benchmark              old allocs     new allocs     delta
BenchmarkSHA256-32     85             85             +0.00%

benchmark              old bytes     new bytes     delta
BenchmarkSHA256-32     29838         23071         -22.68%
```
very yes, neat!

lib/verifier.go
```diff
@@ -0,0 +1,70 @@
...
+		// Check low nibble (last 4 bits)
+		if (b & 0x0F) != 0 {
+			break // Non-zero found in leading nibbles
+		}
+		count++
+		if count >= n {
+			return true
+		}
+	}
+	return count >= n
```
nicoonoclaste, line 71: Since the length is constant and a multiple of 64b, wouldn't that simplify down to something like this?
```go
count := uint32(0)
for _, w := range slices.Chunk(data, 8) {
  d := binary.BigEndian.Uint64(w)
  count += bits.LeadingZeros64(d) / 4
  if d != 0 {
    break
  }
}
return count
```
(using [`slices.Chunk`] to make it more readable, but of course that could be done by a project-local function to avoid that dependency)

[`slices.Chunk`]: https://pkg.go.dev/slices@master#Chunk

PS: Please excuse the approximate Go, I seldom write that language  :sweat_smile:

wasm/pow/argon2id/src/lib.rs
```diff
@@ -0,0 +1,176 @@
...
+/// fact that you cannot easily pass data from host space to WebAssembly space.
+pub static RESULT_HASH: LazyLock<Mutex<[u8; 32]>> = LazyLock::new(|| Mutex::new([0; 32]));
+
+pub static VERIFICATION_HASH: LazyLock<Box<Mutex<[u8; 32]>>> =
+    LazyLock::new(|| Box::new(Mutex::new([0; 32])));
+
+/// Core validation function. Compare each bit in the hash by progressively masking bits until
+/// some are found to not be matching.
+///
+/// There are probably more clever ways to do this, likely involving lookup tables or something
```
aiverson, line 16: You should use the Count Leading Zeros intrinsic.

make-github-pseudonymous-again: - As highlighted in #106, hash validation is simply testing that the given hash is smaller than a given reference depending on difficulty (e.g. difficulty 4 is `hash < 0x00010...` or `hash <= 0x0000f...`). See also:
  - https://github.com/TecharoHQ/anubis/issues/106#issuecomment-2822493966
  - https://github.com/TecharoHQ/anubis/issues/106#issuecomment-2832473776

Issue #242, opened 2025-04-09
OG passthrough: does not send correct Host header


SlyEcho:
The OpenGraph tags caching and forwarding feature from version 1.16.0 is broken on some setups where the Anubis target endpoint does not match the expected hostname of the site.

For example, I have my Mastodon instance set up in Kubernetes, where the main Rails Deployment is behind a ClusterIP Service with the name `mdon-ee-web`, the Ingress is set up on the host `mdon.ee` and it is set to forward to the Service `mdon-ee-anubis` which is exposing the Anubis Deployment. Anubis has been configured with `TARGET=http://mdon-ee-web`.

This works like expected for the regular Anubis functionality. Nginx receives request to IP 152.53.17.226 with the header `Host: mdon.ee`, which it uses to know which rules to use, then it forwards the request to `http://mdon-ee-anubis` inside the cluster keeping the header `Host: mdon.ee`. Anubis forwards the request to the configured target `http://mdon-ee-web` while also keeping the header `Host: mdon.ee`. Mastodon's Rails app sees that the request is for it's configured domain `mdon.ee` and processes the request as usual.

If I turn on the OG passthrough feature, Anubis will do an additional request before showing the challenge page. It makes a regular Go HTTP request to GET the requested URL on the target server. So it makes a request to `http://mdon-ee-web/` with the header `Host: mdon-ee-web`.

```json
{
  "time": "2025-04-07T15:44:48.333436049Z",
  "level": "DEBUG",
  "source": {
    "function": "github.com/TecharoHQ/anubis/internal/ogtags.(*OGTagCache).fetchHTMLDocument",
    "file": "github.com/TecharoHQ/anubis/internal/ogtags/fetch.go",
    "line": 32
  },
  "msg": "og: received non-OK status code",
  "url": "http://mdon-ee-web/@slyecho/113834270598562111",
  "status": 403
}

```

Rails reads the hostname and gives an error, since this hostname is not in its configuration.

```
E, [2025-04-07T15:44:48.332272 #16] ERROR -- : [ActionDispatch::HostAuthorization::DefaultResponseApp] Blocked hosts: mdon-ee-web
```

---

[...]

PR #251, opened 2025-04-11
feat(og): support forwarding headers

jaredallard:
Previously, the `Host` header would not be forwarded which could cause
issues with the application that a request was being sent to. This fixes
that.

Also included is `X-Forwarded-Proto` for the case of being behind a
reverse proxy, such as `Cloudflare`.

This is not a perfect or elegant solution and is meant to solve a few
specific use cases. We should consider something better, such as always
forwarding headers in the future.

Note: I re-ran `npm install` locally which updated `package-lock.json`


jaredallard:
Related issue: https://github.com/TecharoHQ/anubis/issues/242

jaredallard:
Going to leave this as draft because it feels super hacky and I spent only 5 minutes or so on it. If we think this is good enough, can mark as ready, otherwise can debate impl here :eyes:

My main motivator for this right now is that the OG tag impl doesn't work at all for me because my blog requires both of these headers to function properly, which of course `httputil` does handle for us.

JasonLovesDoggo:
> Going to leave this as draft because it feels super hacky and I spent only 5 minutes or so on it. If we think this is good enough, can mark as ready, otherwise can debate impl here :eyes:
> 

My only real concern here is that the cache doesn't take these two headers into account. 

If you send the same a request to your blog with different host headers, you're going to get different HTML back (as well as different OG tags). However, if it's cached it would completely ignore the host header and just see a matching key and return the invalid tags

SlyEcho:
It seems that this is just not enough. I have tested this PR and it does not fix the issue.

Looking at the Go source code for [http.NewRequest](https://go.dev/src/net/http/request.go#:~:text=func%20NewRequest) it seems like the `Host` is a separate field on the Request object and this function will initialize it from the URL given:

```go
func NewRequestWithContext(ctx context.Context, method, url string, body io.Reader) (*Request, error) {
	// ,,,
	u, err := urlpkg.Parse(url)
	// ...
	req := &Request{
		// ...
		Header:     make(Header),
		Host:       u.Host,
	}
	// ...
}
```

The Host value from the Headers collection is actually [ignored](https://go.dev/src/net/http/request.go#:~:text=writeSubset) when the HTTP request is being written:

```go
r.Header.writeSubset(w, reqWriteExcludeHeader, trace)
```

The excluded headers, of course, [defines](https://go.dev/src/net/http/request.go#:~:text=reqWriteExcludeHeader) the `Host` header:

```go
var reqWriteExcludeHeader = map[string]bool{
	"Host":              true, // not in Header map anyway
	// ...
}
```

PR #260, opened 2025-04-13
cmd/containerbuild/main.go: fix docker tag parsing

rayes0:
Minor fix to the docker parsing of tags in the image reference to split at the last colon as per the docker spec[1]. This fixes container builds when the repository already contains an earlier colon, which is common when pushing to a local registry with a port, eg: `localhost:5000/org/name:latest`. The previous would be parsed with the repo as `localhost` and tag as `5000/org/name:latest`, which would cause the build to fail.

[1] https://docs.docker.com/reference/cli/docker/image/tag/#description

Checklist:


Xe (approved):

Xe:
Thanks! I test with an actual registry running on `registry.int.xeserv.us` so I didn't hit this.

PR #266, opened 2025-04-14
Add check endpoint which can be used with nginx' auth_request function

SuperSandro2000:
This is a rebase of https://github.com/Xe/x/pull/694.
So far I did limited testing and I would say with 50% confidence this works as expected. Please give it a careful review.

Also, I am not super happy with the naming of the functions, but I was out of ideas 😓 

This allows to use annubis via nginx' auth_request function.

This is a simplified example extracted from my config:

```
location /.within.website/ {
        proxy_pass http://127.0.0.1:8923;
        auth_request off;
}
location @redirectToAnubis {
        return 307 /.within.website/?redir=$scheme://$host$request_uri;
        auth_request off;
}
location / {
         auth_request /.within.website/x/cmd/anubis/api/check;
         error_page 401 = @redirectToAnubis;
}
```

Checklist:


baconwaifu:
Replicating my comment from the original PR: The only change that would be required to make this also work with traefik would be to return a way to trigger the challenge as part of the response body. Given the mentioned need to return 200 to make bad scrapers actually go away, but traefik needing a non-200 to trigger the hook, how about directly returning the 307 response returned by the sample nginx config instead? (which would probably change `sendStatusOnly` into an enum representing which type of response to send?) 

As to actual code review: there should probably be some simple checks against the redirect-to domain to avoid an open redirect. The easiest no-config option would be to always use host-relative redirects, or sniff the `Host:` header and require the domain to match that.

SuperSandro2000:
nginx needs a 401 or 403 https://nginx.org/en/docs/http/ngx_http_auth_request_module.html

Everything else triggers a 500 IIRC

SuperSandro2000:
I added a commit to allow to configure redirect domains but I have yet to test it with a real world deployment.

Xe (approved):

Xe:
note: this approval means "yeah this looks sensible from a code standpoint", not "I have fully tested this"

Xe:
Hey @SuperSandro2000, I've rebased this on main. I set up an example copy of the configs based on what you provided in [this folder](https://github.com/SuperSandro2000/anubis/tree/auth_request/tests/nginx-external-auth). When I run this (and bypass the TLS warnings for a self-signed cert), I get a redirect to `https://anubis` instead of the final domain `https://nginx.local.cetacean.club`. Is there some low-hanging fruit I'm missing?

ThxAndBye:
I've setup anubis:main and when accessing /x/cmd/anubis/api/check i get a 502 error after the challenge was solved.

As I don't plan to use anubis as a reverse proxy, so I haven't setup any service to proxy.

Wouldn't it make sense if the endpoint would directly return a 401 or 403 code after a successful challenge?

Xe:
Getting this working in a local k3s setup fixed it. Gonna make sure the docs are up to date and ship it. Thanks!

SuperSandro2000:
Thanks for finishing this! I didn't find the time in the last weeks to get back to this. :)

PR #274, opened 2025-04-15
Create Anubis FreeBSD rc.d script

pswilde:
I have added an rc.d script for FreeBSD so Anubis can be run as a daemon.
Typically, this file would be copied to `/usr/local/etc/rc.d` and variables set in `/etc/rc.conf` (i.e. to enable it, and set the environment file).
Tested and running well on my Forgejo Instance - https://forge.notnull.space/

Checklist: 

### DIFF
--- a/run/anubis.freebsd
+++ b/run/anubis.freebsd
@@ -1,0 +1,34 @@
+#!/bin/sh
+
+# PROVIDE: anubis
+# REQUIRE: NETWORKING
+# KEYWORD: shutdown
+
+. /etc/rc.subr
+
+name=anubis
+rcvar=anubis_enable
+
+load_rc_config ${name}
+
+: ${anubis_enable="NO"}
+: ${anubis_user="anubis"}
+: ${anubis_bin="/usr/local/bin/anubis"}
+: ${anubis_environment_file="/etc/anubis.env"}
+
+command=/usr/sbin/daemon
+procname=${anubis_bin}
+pidfile=/var/run/anubis.pid
+logfile=/var/log/anubis.log
+command_args="-c -f -p ${pidfile} -o ${logfile} ${procname}"
+start_precmd=anubis_precmd
+
+anubis_precmd () {
+        export $(xargs < ${anubis_environment_file})
+        if [ ! -f ${logfile} ]; then
+                install -o anubis /dev/null ${logfile}
+        fi
+        install -o anubis /dev/null ${pidfile}
+}
+
+run_rc_command "$1"
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Whitelisted [DuckDuckBot](https://duckduckgo.com/duckduckgo-help-pages/results/duckduckbot/) in botPolicies
 - Improvements to build scripts to make them less independent of the build host
 - Improved the OpenGraph error logging
+- Added FreeBSD rc.d script so can be run as a FreeBSD daemon.
 
 ## v1.16.0
 


run/anubis.freebsd
```diff
@@ -0,0 +1,34 @@
...
+
+: ${anubis_enable="NO"}
+: ${anubis_user="anubis"}
+: ${anubis_bin="/usr/local/bin/anubis"}
+: ${anubis_environment_file="/etc/anubis.env"}
+
+command=/usr/sbin/daemon
+procname=${anubis_bin}
+pidfile=/var/run/anubis.pid
+logfile=/var/log/anubis.log
```
Xe, line 22: Is this where stdout/stderr is written?

pswilde: Yes, correct. 
The daemon command has a `-o` flag which can write the output to a log file.
https://man.freebsd.org/cgi/man.cgi?daemon(8)

Xe (approved):
LGTM, thanks!

run/anubis.freebsd
```diff
@@ -0,0 +1,34 @@
...
+command_args="-c -f -p ${pidfile} -o ${logfile} ${procname}"
+start_precmd=anubis_precmd
+
+anubis_precmd () {
+        export $(xargs < ${anubis_environment_file})
```
dlangille, line 27: From [man 8 rc.subr](https://man.freebsd.org/cgi/man.cgi?query=rc.subr&sektion=8&apropos=0&manpath=FreeBSD+14.2-RELEASE+and+Ports):

```
		   ${name}_env_file
			     A	file  to source	for environmental variables to
			     run command with.	Note: all the variables	 which
			     are  being	 assigned in this file are going to be
			     exported into the environment of command.
```

Might that be useful to you?

pswilde: Useful. 
Do you mean removing the precmd and giving the anubis_env var values via rc.conf? 
personally i prefer using an environment file (portability mainly) so the method created caters for this
however, I see no reason why those who prefer setting environment variables in rc.conf should be disadvantaged and there's no reason why both options can't be available, so yes, very useful indeed.
I suppose it's reasonable to have both in such a way that anubis_env can be set in rc.conf and if an environment file is specified too, the values can be added to anubis_env and continue to work as intended

dlangille: To be clear, I have not tried `anubis_env` yet so it may not work as advertised.

> Useful. Do you mean removing the precmd and giving the anubis_env var values via rc.conf?

No. I think the use of `${name}_env_file` might mean line 27 can be removed - I have not tested this yet. I'm just on my first test deployment of Anubis


> personally i prefer using an environment file (portability mainly) so the method created caters for this however, I see no reason why those who prefer setting environment variables in rc.conf should be disadvantaged and there's no reason why both options can't be available, so yes, very useful indeed. I suppose it's reasonable to have both in such a way that anubis_env can be set in rc.conf and if an environment file is specified too, the values can be added to anubis_env and continue to work as intended

From what I see, this should allow both, for example, the following two line in `/etc/rc.conf` achieve identical results:

e.g.
```
anubis_env='POLICY_FNAME=/usr/local/etc/anubis-freshports.json TARGET=http://localhost:8080'
anubis_env_file="/usr/local/etc/anubis.env"
```

where:

```
% cat /usr/local/etc/anubis.env
POLICY_FNAME=/usr/local/etc/anubis-freshports.json
TARGET=http://localhost:8080
```

Users can use either or a combination of both. Testing would reveal if specifying the same VAR in both places indicates a precedence (e.g. that one approach always overrides the other).

At present, I am using the `rc.d` script from the FreeBSD package - [anubis.in](https://cgit.freebsd.org/ports/tree/www/go-anubis/files/anubis.in) - it passes in `anubis_args` to `anubis`. For example, I have these settings:

```
 [11:24 test-nginx01 dvl ~] % grep anubis /etc/rc.conf
anubis_enable="YES"
anubis_args="-target http://localhost:8080 -bind :8923 -difficulty 2 -cookie-domain=freshports.org -policy-fname=/usr/local/etc/anubis-freshports.json"
```

When started, I get this:

```
[11:24 test-nginx01 dvl ~] % ps auwwx | grep anubis  
www        20041  0.0  0.0   13856  2320  -  IsJ  19:52    0:00.00 daemon: /usr/local/sbin/anubis[20042] (daemon)
www        20042  0.0  0.0 1306896 24652  -  IJ   19:52    0:10.60 /usr/local/sbin/anubis -target http://localhost:8080 -bind :8923 -difficulty 2 -cookie-domain=freshports.org -policy-fname=/usr/local/etc/anubis-freshports.json
```

pswilde: Great, thanks. This makes sense. I'll do some testing.
Yep, I'm aware of the ports version - and the creator of that one and I have already spoken. That one also has some better daemon usage so may be worth migrating a bit of that in too.

pswilde: This is good stuff. I've tested and each variable can work with the other
`${name}_env_file` gets read first, and environment variables configured. Followed by `${name}_env`, which only overrides matching env var key values. 
`${name}_args` is of course more anubis specific as to what priority values have over environment variables, but this all appears to be fine so values in `${name}_args` override matching keys in both environment vars.

Looking nice!
I've made some adjustments and put them in PR #427

PR #321, opened 2025-04-22
feat: enable loading config fragments

Xe:
Checklist:

- [x] Changelog entry
- [x] Split out config file into fragments
- [x] Add an `import` or `uses` stanza to the config file format, test that
- [x] Adapt example configs to use that
- [x] Documentation on imports
- [x] Testing


JasonLovesDoggo:
Once this is merged i'll write a PR to auto update those CIDR ranges

PR #347, opened 2025-04-23
fix(lib): use correct URL for path checker in PassChallenge

compilade:
This makes `PassChallenge` use the path from the url given in the `redir` form value of `/.within.website/x/cmd/anubis/api/pass-challenge`.

Otherwise, `r.URL.Path` was always `/.within.website/x/cmd/anubis/api/pass-challenge` and this didn't match the path checker rules correctly, which caused a failure when the difficulty mismatched (a 403 error page because the challenge doesn't match).

An easy way to reproduce the problem is with a `botPolicy.json` which uses a different difficulty on a path than another rule which would match if the path did not:

```yaml
{
  "bots": [
    {
      "name": "custom-difficulty",
      "path_regex": "^/$",
      "action": "CHALLENGE",
      "challenge": {
        "difficulty": 3,
        "report_as": 4
      }
    },
    {
      "name": "generic-browser",
      "user_agent_regex": "Mozilla",
      "action": "CHALLENGE",
      "challenge": {
        "difficulty": 4,
        "report_as": 4
      }
    }
  ]
}
```

and then trying to access a page affected by the `path_regex` (in this case `/`) with a browser.

Checklist:

  - I'm not quite sure how to write the test, especially since also `MakeChallenge` has no way to get the actual URL path to match `path_regex` rules. I guess this should instead be tested with the playwright integration tests?

### DIFF
--- a/lib/anubis.go
+++ b/lib/anubis.go
@@ -12,6 +12,7 @@ import (
 	"math"
 	"net"
 	"net/http"
+	"net/url"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


Xe (approved):

Xe:
Good catch!

Xe:
Ah can you take a look at the docs on commit signing? SSH key signing is fine: https://anubis.techaro.lol/docs/developer/signed-commits

PR #328, opened 2025-04-22
feat(cmd/anubis): compute full XFF header

acuteaura:
Checklist:

- [x] Added a description of the changes to the `[Unreleased]` section of docs/docs/CHANGELOG.md
- [ ] Added test cases to [the relevant parts of the codebase](https://anubis.techaro.lol/docs/developer/code-quality)
- [ ] Ran integration tests `npm run test:integration` (unsupported on Windows, please use WSL)

PR #350, opened 2025-04-24
feat: more elaborate XFF compute

### LINKED ISSUE #344: X-Forwarded-For broken after #328
I'm guessing https://github.com/TecharoHQ/anubis/pull/328 that was recently merged has broken it for me:

`{"time":"2025-04-23T19:14:30.083671666+01:00","level":"ERROR","source":{"function":"github.com/TecharoHQ/anubis/lib.(*Server).MaybeReverseProxy","file":"g
ithub.com/TecharoHQ/anubis/lib/anubis.go","line":201},"msg":"check failed","user_agent":"Mozilla/5.0 (iPhone; CPU iPhone OS 18_4_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.
4 Mobile/15E148 Safari/604.1","accept_language":"en-GB,en;q=0.9","priority":"u=0, i","x-forwarded-for":",,169.254.1.2","x-real-ip":"","err":"[misconfiguration] X-Real-Ip header is not set"}`

I use Caddy in one container to talk to Anubis in another container and on rootless Podman that maps to a source IP of 169.254.1.2. This also relates to https://github.com/TecharoHQ/anubis/issues/103 as XFF doesn't work for local IPs if X-Real-Ip isn't set. I can't reliably use X-Real-Ip in Caddy because it doesn't work with HTTP/3 0-RTT Early Data. I worked around that by using Caddy to not send local IP ranges to Anubis at all, but now that's broken again.

acuteaura:
#328 followup

now featuring configuration and
defaults that shouldn't break most
setups.

fixes #344


acuteaura:
I plan on writing a nice test suite for `computeXFFHeader` when I find some more time...

Xe:
I added a small test environment for your fixes, and it passes that test environment! I think this is good to ship.

Xe (approved):

acuteaura:
Ah, I was gonna write a `go test` unit test, being that I extracted the function that has all the logic, but I can do that in a followup

Xe:
That works too! You nerd sniped me into making better functional testing infrastructure for cases like this :)

internal/headers.go
```diff
@@ -115,19 +116,24 @@ func XForwardedForUpdate(next http.Handler) http.Handler {
...
 	parsedRemoteIP, err := netip.ParseAddr(remoteIP)
 	if err != nil {
-		return "", fmt.Errorf("unable to parse remote IP: %w", err)
+		return "", fmt.Errorf("%w: %w", ErrCantParseRemoteIP, err)
 	}
 
 	origForwardedList := make([]string, 0, 4)
 	if origXFFHeader != "" {
-		origForwardedList = strings.Split(",", origXFFHeader)
+		origForwardedList = strings.Split(origXFFHeader, ",")
```
Xe, line 136: This subtle bugger was causing some issues!

internal/headers.go
```diff
@@ -177,7 +177,9 @@ func computeXFFHeader(remoteAddr string, origXFFHeader string, pref XFFComputePr
 		xffHeaderString = ""
 	}
 	if pref.Flatten {
-		xffHeaderString = forwardedList[len(forwardedList)-1]
+		if len(forwardedList) != 0 {
```
acuteaura: I forgot to add a return to the branch above, that'll also fix it.

Xe:
Thank you so much for this fix!

PR #389, opened 2025-04-28
feat: make authorization cookie default expiration time customizable

### LINKED ISSUE #272: Allow to change cookie expiration time
The expiration time of the cookie/JWT is hard-coded, it would be very useful if admins could change it through configuration.

Maybe allowing it to be changed in the bot policies could be useful too 🤔

Nydauron:
Closes #272. Adds `--cookie-expiration-time` flag and env variable to allow for customizing the expiration date of authorization cookies. It takes in a duration similar to `--og-expiry-time` representing the amount of time til the cookie expires.

Checklist:


lib/anubis.go
compilade, line 1: Should the `24 * 7 * time.Hour` in `challengeFor` also be configurable through this? Otherwise, the challenge will basically be the same per client for the week and the previous solution can be re-used even if the cookie expires quickly.

However, whenever the time rounds in `challengeFor`, the challenge given is not necessarily the same as the challenge it's verified against by the time it's solved. (e.g. if the expiration time is an hour, then at each hour boundary that problem can occur)

But having the capability to make this happen more often will likely result in finding a solution.

anubis.go
```diff
@@ -11,6 +13,9 @@ var Version = "devel"
 // access.
 const CookieName = "within.website-x-cmd-anubis-auth"
 
+// CookieExpirationDuration is the amount of time before the cookie/JWT expires.
+const CookieExpirationDuration = 7 * 24 * time.Hour
```
JasonLovesDoggo: Given that this isn't the actual value used. I would remove this and just rely on the flag

Xe (approved):

anubis.go
JasonLovesDoggo: Or rename to default cookie expiry time

lib/anubis.go
Nydauron: Hmm, I'm getting what you are putting down. Currently, the only way I can think of entirely mitigating this involves extra state on the backend (e.g. making the challenge based on a random value which is stored server-side).

Another alternative would be to accept a solution before a certain amount time has elapsed (exact time used in the creation of the challenge and the challenge would have to be put into a JWT with a set expiration time and this JWT would have to be sent to the client and sent with the solution on challenge submission), but currently, solving time varies quite a lot between attempts and different devices. While not a perfect solution, the timeout can be set to a very high upper bound and scales exponentially with difficulty (e.g. 16^x), with minimal impact to real users. Based on the numbers we've been gathering, 8-10 mins after issuance when difficulty = 4 seems to be rather generous (99%-ile of clients take about 30 seconds, 99.6%-ile taking around 87 seconds).

Xe:
I rebased this on main and did manual testing by setting the lifetime absurdly low (30s). Thanks! This is good to ship.

PR #393, opened 2025-04-29
feat(config): add ability to customize HTTP status codes Anubis returns

### LINKED ISSUE #355: Considering / Configuring Anubis response code
Hi there,

we are currently testing a deployment of Anubis for Codeberg.org. However, we expect that the response code of error 200 results in a lot of unexpected breakage and support overhead, because debugging is harder.

For example, tools might correctly report connection issues if they receive an invalid response code. Serving data with "200 OK" could result in arbitrary issues like parsing issues, app crashes and much more. Even the Forgejo-native "webhook" test feature only checks for the 200 response code and would mark delivery as successful, although it might have been trapped by anubis.

I understand the motivation to trick crawlers into thinking they were successful, but this is probably not suitable for deployments that have legitimate and unknown automated traffic.

Xe:
Closes #355 

Checklist:

### DIFF
--- a/data/botPolicies.yaml
+++ b/data/botPolicies.yaml
@@ -48,3 +48,11 @@ bots:
   action: CHALLENGE
 
 dnsbl: false
+
+# By default, send HTTP 200 back to clients that either get issued a challenge
+# or a denial. This seems weird, but this is load-bearing due to the fact that
+# the most aggressive scraper bots seem to really really want an HTTP 200 and
+# will stop sending requests once they get it.
+status_codes:
+  CHALLENGE: 200
+  DENY: 200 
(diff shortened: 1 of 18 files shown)


MatteoJoliveau:
Random idea: could it be interesting to allow customisation of the status code per bot?
For instance, we could return 200 for all bots but override it to 403 for a specific bot known to respect it.

```json
{
  "bots": [
    {
      "name": "nice-bot",
      "user_agent_regex": "NiceBot",
      "action": "CHALLENGE",
      "status_codes": {
        "CHALLENGE": 403,
        "DENY": 403
      }
    },
    {
      "name": "everything",
      "user_agent_regex": ".*",
      "action": "DENY"
    }
  ],
  "status_codes": {
    "CHALLENGE": 200,
    "DENY": 200
  }
}
```

Xe:
I like that idea, but let's circle back after this ships because that requires some annoying refactoring.

PR #412, opened 2025-05-01
fix(web): Avoid timestamping main.mjs.gz

heftig:
Improves reproducibility.


Xe (approved):
Note for review: the change to add `-n` to the `gzip` call disables gzip embedding the original filename and timestamp in gzip's metadata. This made every build of `anubis` have different static file contents, meaning that the binary was not reproducible (in a reproducible build, the same source input gets you the same source output).

PR #427, opened 2025-05-02
Overhaul anubis.freebsd

pswilde:
Some changes here to reflect the discussion in pull request 274 regarding the `anubis_env`, `anubis_env_file` and `anubis_args` variables. At the risk of improving personal choices in configuration with a minor amount more complexity, this new script now allows for the use of all three of these, together, with no interference between them all  i.e. 
- if `anubis_env_file` is set, environment variables will be taken from this file
- if `anubis_env` is set, environment variables will be taken from this string of variables, and override matching variables set in `anubis_env_file`
- if `anubis_args` is set, runtime parameters will be taken from this string and override matching ones in both `anubis_env_file` and `anubis_env`

Thanks to @dlangille for the advice with this.

Checklist:

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Add `check-spelling` for spell checking
 - Add `--target-insecure-skip-verify` flag/envvar to allow Anubis to hit a self-signed HTTPS backend.
+- Minor adjustments to FreeBSD rc.d script to allow for more flexible configuration.
 
 ## v1.18.0: Varis zos Galvus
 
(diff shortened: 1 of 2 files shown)


dch:
Last I checked things like env vars don’t get correctly passed through when using daemon, which is why the rc script in ports uses anubis_args instead of the more obvious anubis_flags.

if these changes above work I should adopt them in the port … perhaps we should keep the rc script in one place rather than two?

pswilde:
Yep, that's fair. I wrote the original before I knew of your ports version, so I agree this is largely unecessary now.
I am running with env vars via daemon successfully in production however, and on another service with all three`env_file`, `env` and `args` with overrides without issue. So perhaps whatever didn't work before works now, or I've just been lucky 😄

Anyway, very happy for ports to be the mainline, as it should be. Perhaps a note in docs to point people that way?

Xe:
I'd be more than happy to have the ports version be the way we suggest for FreeBSD. Just make sure it's up to date!

dch:
I updated to 1.18.0 today, thanks heaps @Xe for the `pre-baked`
target, it made it far too easy!

Issue #428, opened 2025-05-02
Don't check if cookies are blocked. Just block


pato-pan:
Don't refresh the page and don't try to verify the user if they are blocking cookies. Just ask them to enable cookies.

It has the following problems
1. Minimal cost to bandwidth
2. Minimal cost on performance on pretty old devices or a virtual machine or if you have too many problems

I personally just find it mildly annoying to see the page refresh over and over. This thing can be smarter like cloudflare and just don't do that if you have them blocked

Xe:
How am I meant to track the state of the repeated re-challenges if you don't store cookies?

Xe:
If you disable cookies, it breaks Anubis, and probably most of the internet. Closing as not planned.

pato-pan:
This should be marked as solved in case someone finds it

PR #437, opened 2025-05-04
Fail immediately if pass-challenge is sent without cookie

afontenot:
This is a straightforward way to test for cookie support in the user
agent. It doesn't change the behavior for any other user agents, but
browsers that have cookies disabled will show an error rather than
refreshing endlessly.

This fixes a common user complaint - #428 and also discussed on HN:
https://news.ycombinator.com/item?id=43865151

Most users don't block cookies entirely in their browser, but this
would help the few that do, and for users selectively enabling cookies
for certain sites, this will make allow-listing Anubis fronted sites
easier because the page will not be redirecting while they're trying
to allow the cookie to be set.

Checklist:

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Change import syntax to allow multi-level imports
 - Changed the startup logging to use JSON formatting as all the other logs do.
 - Added the ability to do [expression matching with CEL](./admin/configuration/expressions.mdx)
+- Changed pass-challenge check to fail immediately if no cookie is set.
 
 ## v1.17.1: Asahi sas Brutus: Echo 1
 
--- a/lib/anubis.go
+++ b/lib/anubis.go
@@ -268,6 +268,14 @@ func (s *Server) PassChallenge(w http.ResponseWriter, r *http.Request) {
 	}
 	lg = lg.With("check_result", cr)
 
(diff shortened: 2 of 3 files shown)


afontenot:
This was a proof of concept I wrote to solve an annoyance of mine in 10 minutes. I will very happily add / run tests if the basic concept behind this change is okayed, and make other changes as requested.

afontenot:
I tested this with [nuMatrix](https://codeberg.org/arek/nuMatrix), the extension I use to selectively block requests, and not only does this work really well, it even *immediately* passes the check from the error page if you allow cookies in nuMatrix and then retry.

afontenot:
It might make sense to refactor `ClearCookie` to `ResetCookie`, since the cookie is (briefly) left intact after this change.

afontenot:
> i'd probably use [navigator.cookieEnabled](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/cookieEnabled) in javascript instead (or additionally), so the user doesn't need to wait (duration of PoW) to get the no cookie error.

You're right, this would make sense as the first line of defense. I think most cookie blocking extensions work by filtering outgoing headers, though, so something like this is still needed.

Xe:
I took a stab at implementing this in the list of client quirks (or whatever we're calling it): https://github.com/TecharoHQ/anubis/pull/438.

Xe:
I'll take a stab at making an implementation of the cookie test (with a separate cookie name because otherwise I don't trust most of the rest of the stack handling this properly), but blocking first-party cookies makes this logistically difficult.

Xe:
Superceded by #501

PR #471, opened 2025-05-08
fix(js): use pure JS SHA256 library, refactor

### LINKED ISSUE #458: Firefox, 26.934 seconds to solve level 4 challenge
In my experience, Firefox had been very slow to solve Anubis CAPTCHAs, despite running on a modern CPU, compounded by my preference to erase non-whitelisted cookies on restart. That means I have to see multiple Anubis challenges on multiple websites on most days. The last number came from GNOME GitLab's 27 seconds , being second to Arch Wiki's reproducible minute of waiting. The shortest time was <300 ms on SourceHut, though I recall that was a <4 level challenge, so that would be expected. When Anubis first appeared, Firefox could solve its level 4 challenges in about a second. I didn't observe similar slowdowns on Android Chromium.

I wanted to paste the stats from that most recent encounter, but it redirected me to the original page while idle, so I attempted reproducing it on https://anubis.techaro.lol/.

```
Difficulty: 4, <Avg> Speed: about 4~5 kH/s.
Took 14184ms, 73553 iterations.
```

I searched for any similar issues, but couldn't find any. Is this the expected behavior?

Xe:
Closes #458

Additionally, I made a horrifying discovery: Firefox seems to actively hinder performance if you are using more than one Worker per page. It does not spread the load out across cores like I expected. Instead it seems to make that one Worker thrash and have to constantly context switch, which caused a lot of slowdown.

The benchmarks in #155 continue to be the best contribution ever made to Anubis. What clued me into there being a problem here was the fact that the "slow" algorithm was faster than the "fast" algorithm on my laptop. This made no intuitive sense to me so I dug further.

Either way I think this is a Firefox bug at its core, but for now we have to work around it by doing the hacky terrible thing that I hate.

I also swapped the SHA256 operations to @aws-crypto/sha256-js on the advice of a trusted cryptography expert. I don't know what performance differences this makes, but I'm getting 150-225 kilohashes per second, which is pretty dang good.

Checklist:


Taywee:
I noticed your `algorithms/fast.mjs` is still pointing at the `slow.mjs` worker file.  Might that be the cause of the discovery?  On Linux Firefox, I'm still getting far better performance with fast than slow.

Taywee:
Using the benchmarking code, I get about 180 kH/s on the slow, 275 on fast with the `return 1` in `determineThreadCount`, and about 800 with the firefox section commented out.

Interestingly, Chromium is now significantly slower than Firefox for me with this code.

DiegoPino:
Hi. Might be just me (stuck old cookie maybe?) but after this commit I got "Failed to calculate challenge: undefined" on Firefox. I will re-test tomorrow on a fresh container. Just in case I am not the only one.

PR #478, opened 2025-05-09
Please call it stonks‼️

IyaadArshad:
Star history sounded boring, therefore, I have renamed it to stonks‼️


Xe:
Out of scope for Anubis

PR #501, opened 2025-05-13
feat(lib): ensure that clients store cookies

Xe:
If a client is misconfigured and does not store cookies, then they can get into a proof of work death spiral with Anubis. This fixes the problem by setting a test cookie whenever the user gets hit with a challenge page. If the test cookie is not there at challenge pass time, then they are blocked. Administrators will also get a log message explaining that the user intentionally broke cookie support and that this behavior is not an Anubis bug.

Additionally, this ensures that clients being shown a challenge support gzip-compressed responses by showing the challenge page at gzip level 1. This level is intentionally chosen in order to minimize system impacts.

The ClearCookie function is made more generic to account for cookie names as an argument. A correlating SetCookie function was also added to make it easier to set cookies.

Checklist:

### DIFF
--- a/lib/anubis.go
+++ b/lib/anubis.go
@@ -316,7 +316,7 @@ func (s *Server) PassChallenge(w http.ResponseWriter, r *http.Request) {
 		s.ClearCookie(w, s.cookieName)
 		s.ClearCookie(w, anubis.TestCookieName)
 		lg.Warn("user has cookies disabled, this is not an anubis bug")
-		s.respondWithError(w, r, "Your browser is configured to disable cookies. Anubis requires cookies for the legitimate interest of making sure you are a valid client. Please enable cookies for this domain.")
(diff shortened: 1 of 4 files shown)


lib/anubis.go
```diff
@@ -310,19 +312,29 @@ func (s *Server) PassChallenge(w http.ResponseWriter, r *http.Request) {
 
 	challenge := s.challengeFor(r, rule.Challenge.Difficulty)
 
+	if _, err := r.Cookie(anubis.TestCookieName); err == http.ErrNoCookie {
+		s.ClearCookie(w, s.cookieName)
+		s.ClearCookie(w, anubis.TestCookieName)
+		lg.Warn("user has cookies disabled, this is not an anubis bug")
+		s.respondWithError(w, r, "Your browser is configured to disable cookies. Anubis requires cookies for the legitimate interest of making sure you are a valid client. Please enable cookies for this domain")
```
lmfont, line 319: nit: missing final `.`

Xe: respondWithError adds it

Xe: Maybe this is confusing behavior that should be removed lol

PR #510, opened 2025-05-16
fix(systemd): add RuntimeDirectory

### LINKED ISSUE #508: Sockets need a RuntimeDirectory to work properly
Hi,

I ran into quite a bit of difficulty getting Anubis to work properly when I tried installing it from the .deb package and setting it up to use a UNIX socket right off the bat. At first I couldn't figure out why it kept entering the failed state (there was no additional information provided by `systemctl status`), but I eventually found `listen unix /run/anubis.sock: bind: permission denied` messages in `journalctl`.

I found some limited success in getting it working temporarily by trying different directories and messing with tmpfiles.d, but that caused problems when I attempted to reboot the container because the anubis user does not exist until systemd creates it on the fly with `DynamicUser=yes` in the unit file.

The best solution I found was to add `RuntimeDirectory=anubis` to the service unit file and set the socket path to `/run/anubis/anubis.sock`. I propose this be added to the base unit file since attempting to open a socket directly in /run apparently doesn't work, and having a runtime directory to organize your Anubis sockets will probably be useful anyhow if you are running several instances.

Xe:
Closes #508

Notably, this makes the Anubis RuntimeDirectory `/run/anubis` instead of `/run/anubis/%i`. I don't think this is confusing in practice, but it could come up as weird.

Checklist:

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updated the nonce value in the challenge JWT cookie to be a string instead of a number
 - Rename cookies in response to user feedback
 - Ensure cookie renaming is consistent across configuration options
+- Add `RuntimeDirectory` to systemd unit settings so native packages can listen over unix sockets
 
 ## v1.18.0: Varis zos Galvus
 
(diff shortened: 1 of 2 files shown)


Xe:
oh no, do I need to do `/run/anubis/%i` then?

Issue #569, opened 2025-05-27
No CIDR support in expressions, no "match-all" in regular rules


thess:
Thanks for this add-on service - Our site(s) are returning to somewhat normal responsiveness. This problem can be solved with Nginx rules and maps but it would be nice to use Anubis for a solution.

Neither of the following rules work as hoped.
```
- name: deny-mirrors
  action: DENY
  user_agent_regex: (?i)wget|lftp
  remote_addresses: [ "123.42.60.0/24", "5.62.42.5/32" ]
```
-- or --
```
- name: deny-mirrors
  action: DENY
  expression:
    all:
      - userAgent.matches("(?i)wget|lftp")
      - remoteAddress in [ "123.42.60.0/24", "5.62.42.5/32" ]
```
The first rule doesn't work when the UA is "Wget" for example. The pattern matcher quits on first found item and the IP is not considered (re-ordering doesn't help). The second is not possible since the 'list' of addresses is just strings and no CIDR IP matching is possible (AFAICT).

PR #593, opened 2025-06-01
feat(expressions): Add function remoteAddrInList to CEL expressions.

thess:
Allows bot policy rule to check if remote client address falls within a list of CIDR ranges

For example:

```yaml
  - name: deny-unwanted-scrapers
    action: DENY
    expression:
      all:
        - userAgent.matches("(?i)Wget|lftp")
        - remoteAddrInList(remoteAddress, [ "5.107.206.0/24", "2603:c022:8002:6200::8888/128" ])
```

FWIW - this is my attempt to address #569. I'm not a Golang programmer but was willing to give it a try (_hey - code is code_) to see if I could resolve my own issue. This function does seem to work OK - I have tested it live but would like to figure out an integration test. `Playwright_test` looks like a place to start. Probably needs re-factoring too.

Checklist:


Xe:
Thank you so much for this! I'm going to take a look tomorrow once I wake up.

lib/policy/expressions/environment.go
```diff
@@ -2,13 +2,50 @@ package expressions
...
+			continue
+		}
+
+		_, rng, err := net.ParseCIDR(string(cidr))
+		if err != nil {
+			return false, fmt.Errorf("remoteAddrInList: address %s CIDR parse error: %w", cidr, err)
+		}
+
+		ranger.Insert(cidranger.NewBasicRangerEntry(*rng))
+	}
```
Xe: Is there a way to cache this? Feels like this is a poster child for data that we don't have to parse every time.

Xe (approved):
I love this!

lib/policy/expressions/environment.go
thess: I totally agree re-parsing the IP list over-and-over seems a bit much. I really haven't looked too deeply into the cel-go compilation and execution process. It appears that something could be done with the intermediate structure (ast?) but that was beyond what I was trying to accomplish at the moment. Finding any documentation on manipulating the `ast` structures seems hard to find. I'm still trying to figure out how 'NewCELChecker' builds the rule for execution. The current 'expression' implementation doesn't seem to support any nesting and can only have one of a `single expression`, `all` or `any` block. The one example in the doc that shows an `all:` rule not inside an `expression` seems to be in error. https://anubis.techaro.lol/docs/admin/configuration/expressions

```yaml
- name: challenge-wiki-history-page
  action: CHALLENGE
  all:
    - 'path == "/index.php"'
    - '"title" in query'
    - '"action" in query'
    - 'query["action"] == "history"
``` 

My original thought (and test-case) was to make bot rules 'match-all' rather than 'first-match' but that seemed too dramatic a change to the current bot rule semantics. Perhaps a new rule type which implies 'match-all' would be the easiest since it leverages existing mechanisms.

JasonLovesDoggo: I would simply move the codeblock of L24 to L39 into a separate function which is cached. The IP ranges should be relatively consistent with `remoteAddr` being the inconstant one. If you want some help on this PR feel free to give me a shout

thess: Sure, I would greatly appreciate the help -- Golang is not my thing.

I have been experimenting (proto fully functional) with non-expression rules and adding a `match-all` tag to rule definitions. Go doesn't easily lend itself to function overloading so I had to tweak the `Checker` object a bit. If you would prefer to go (_haha_) with a different solution to the same problem I can submit another PR.

JasonLovesDoggo: I would keep this PR at the scope that it's at now. 
If you could give me push access to your fork or if Xe adds me as a maintainer here, I can push the caching patch. 

Alternatively if you want to write the code yourself, shoot me a DM on Discord `personality.json`. If there is some other platform you'd prefer just LMK

thess: Rather than grant write access to my forked repo - would you mind posting a PR to my branch and I will merge and update this PR. Thanks!

JasonLovesDoggo: :D Sounds good. Just did some tests and we end up with a 10.9x performance boost in benchmarks. 

https://github.com/thess/anubis/pull/1

thess:
Updated to current trunk (again)
@Xe -- Are we going to merge this for 1.20 ?

JasonLovesDoggo:
> Updated to current trunk (again) @Xe -- Are we going to merge this for 1.20 ?

Seems like time flies. Lets get this merged in :D 

Given that it's a something new and the amount of changes since Xe approved it. I'm going to leave @Xe to merge it in.

Issue #564, opened 2025-05-26
"invalid response." after "Success!" in Chromium


CyberShadow:
Hi, thank you very much for Anubis!

I've just set it up for two websites on the same server:

- https://wiki.dlang.org/ - works perfectly.
- https://dsource.org/ :
  - In Firefox: works perfectly.
  - In Chomium: In a new private window, I see, in quick succession:
    1. "Making sure you're not a bot"
    2. "Success!"
    3. "Oh noes!" / "invalid response"
       - On this page, nothing happens if I click "Try again", but clicking "Go home" results in a "Success!" and successful website load.

Any idea what could be the problem?

Installed from `anubis_1.18.0_amd64.deb`.
Browser versions:
- Firefox 136.0.3
- Chromium 134.0.6998.165

mikepayne02:
openwrt.org started working again with Chrome 136.0.7103.114 today, strange.

Xe:
Do you have a dual stack internet connection?

CyberShadow:
> Do you have a dual stack internet connection?

Yes, that must be it. Firefox is using IPv4 while Chromium is using IPv6.

mikepayne02:
Confirmed, I was using a dual-stack network yesterday and now on an IPv4 only one it works.

me-cfs:
I am getting this error as well on Safari ios

Xe:
I'm going to implement a different challenge string creation method when I get off work today. I'm going to make something that operates off of client metadata without the remote IP address to see if that causes a reduction in the false rejection rate. I've been doing research on a fairly large dataset of HTTP requests sent to my infrastructure and I think I've found some patterns that can create a somewhat unique fingerprint without JavaScript in the mix.

```text
db> select count(*) as requests from cloud.requests;
```

| requests |
|---------:|
| 300655   |

I plan to do this in a few phases:

- [x] Document the `thr1` (Techaro HTTP Request Fingerprinting version 1) spec on my git server
- [ ] Make a pass at implementation, see what the collision rate is for requests from different ASNs
- [ ] Push as a PR to Anubis and deploy to a few sites to get a feeling for what it's like in practice

CyberShadow:
Thank you, may I suggest to perhaps look into fixing the "Try again" button in this situation, as that seems like a low hanging fruit for UX improvement.

Xe:
Fix it how? I'm not aware of it not working.

CyberShadow:
[...]

PR #616, opened 2025-06-05
docs: add THR1 spec

Xe:
Ref #564

THR1 is a novel HTTP client fingerprinting method that operates on the headers that a HTTP client provides. It works under the constraints that Go's standard library HTTP server sets. This is intended to work with another fingerprinting method such as JA4 TLS client fingerprinting.

This PR introduces the documentation for this HTTP client fingerprinting spec based on patterns I have seen in the real world.

A future PR will introduce robust support for THR1 and replace the existing HTTP client fingerprinting with THR1 and JA4 if the upstream reverse proxy supports it.

### DIFF
--- a/docs/docs/developer/thr1.mdx
+++ b/docs/docs/developer/thr1.mdx
@@ -1,0 +1,187 @@
+# Techaro HTTP Request Fingerprinting Version 1
+
+The naïve way to identify HTTP clients is to use the HTTP User-Agent string as a signal. In an ideal world, this would give you a perfect view of what clients are connecting to your server. We do not live in that ideal world. As such, we need an alternative method that can scale to the world we have.
+
+## Prior Art
+
+The biggest source of prior art is [FoxIO's JA4H fingerprinting method](https://github.com/FoxIO-LLC/ja4/blob/main/technical_details/JA4H.md). This is fine, but there's a problem with it in the real world: Go doesn't allow you to observe the order headers arrived in. As Anubis is written in Go and I don't feel like boiling the HTTP server ocean today, there needs to be an alternative.
+
+## THR1
+
+The fingerprint consists of four concatenated components:
+
+```text
+<thr1_head>_<thr1_lang>_<thr1_sec>_<thr1_all>
+```
+
+Example:
+
+```text
+get20cr1004_enca-d6b272e5b_sec-a9649072c_2a347fcf7
+```
+
+Each component is described below:
+
+### `thr1_head`
+
+Overall request summary of method, protocol, and header counts:
+
+- First three letters of the HTTP method, lowercased (e.g. get, pos).
+- HTTP protocol version formatted in two digits (`10` for HTTP/1.0, `11` for HTTP/1.1, `20` for HTTP/2, `30` for HTTP/3 etc.).
+- Single letter indicating if the request has cookies: `c` if present, `n` if not.
+- Single letter indicating Referer header presence: `r` if present, `n` if absent.
+- Number of HTTP headers sent by the client, zero-padded to two digits (e.g. `10`).
+- Number of `Sec-*` headers sent by the client, zero-padded to two digits (e.g. `04`).
+
+Example:
+
+```text
+get20cr1004
+```
+
+### `thr1_lang`
+
+`Accept-Language` header details.
+
+- If no `Accept-Language` header is set, then:
+
+  ```
+  -000000000
+  ```
+
+- Otherwise:
+  - The first 4 alphanumeric characters of the header value (lowercased, right-padded with `0` to length 4), e.g. `enca`.
+  - The first 9 hex characters of the SHA-256 hash of the full `Accept-Language` header value.
+
+Example:
+
+```
+enca-d6b272e5b
+```
+
+### `thr1_sec`
+
+Details about the `Sec-*` headers sent by the client.
+
+```
+thr1_sec = "sec-" + HASH9
+```
+
+Where:
+
+- Collect **all headers whose names start with `sec-` (case-insensitive)**.
+- For each header:
+
+  1. Normalize the header name by lowercasing.
+  2. If the header is one of the `Sec-CH-UA` family:
+
+     - `Sec-CH-UA`
+     - `Sec-CH-UA-Mobile`
+     - `Sec-CH-UA-Platform`
+     - `Sec-CH-UA-Platform-Version`
+     - `Sec-CH-UA-Model`
+     - `Sec-CH-UA-Full-Version`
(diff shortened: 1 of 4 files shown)


docs/docs/developer/thr1.mdx
```diff
@@ -0,0 +1,187 @@
+# Techaro HTTP Request Fingerprinting Version 1
+
+The naïve way to identify HTTP clients is to use the HTTP User-Agent string as a signal. In an ideal world, this would give you a perfect view of what clients are connecting to your server. We do not live in that ideal world. As such, we need an alternative method that can scale to the world we have.
```
ryanccn, line 3: The purpose of the User-Agent header has never been to identify individual clients/users. It is for identifying the user *agent*, which is just the product and version that the user happens to be using. In fact, RFC 7231 explicitly states that

> A user agent SHOULD NOT generate a User-Agent field containing needlessly fine-grained detail and SHOULD limit the addition of subproducts by third parties. Overly long and detailed User-Agent field values increase request latency and the risk of a user being identified against their wishes ("fingerprinting").

docs/docs/developer/thr1.mdx
```diff
@@ -0,0 +1,187 @@
...
+
+Each component is described below:
+
+### `thr1_head`
+
+Overall request summary of method, protocol, and header counts:
+
+- First three letters of the HTTP method, lowercased (e.g. get, pos).
+- HTTP protocol version formatted in two digits (`10` for HTTP/1.0, `11` for HTTP/1.1, `20` for HTTP/2, `30` for HTTP/3 etc.).
+- Single letter indicating if the request has cookies: `c` if present, `n` if not.
```
ryanccn: This is a confusing choice. Cookies are mutable state on the part of the client, and the presence or absence of cookies should not be a factor in a fingerprint. For example, Anubis itself sets a cookie when a challenge has been solved by the client. This would change the fingerprint from `n` to `c` if no cookies were present before, thereby immediately invalidating the previously solved challenge.

docs/docs/developer/thr1.mdx
```diff
@@ -0,0 +1,187 @@
...
+- Single letter indicating Referer header presence: `r` if present, `n` if absent.
```
ryanccn: Including `Referer` is also somewhat questionable, since this would result in different fingerprints depending on whether the user directly navigates to a website or opens it from an internal or external link. This behavior does not particularly matter for bots that repeatedly make requests with the same headers, but for human users I think this would cause a few problems.

docs/docs/developer/thr1.mdx
```diff
@@ -0,0 +1,187 @@
...
+- Number of HTTP headers sent by the client, zero-padded to two digits (e.g. `10`).
```
ryanccn, line 32: This faces the same problem as the previous review comment. A human client interacting with the user agent could possibly generate a variety of header combinations, and for them a frequently changing fingerprint would be something of a nuisance.

I noticed that the inclusion of cookies and referrer in the fingerprint is borrowed from JA4H, and I'm not sure I completely understand why JA4H does include these, but I would think that identifying malicious patterns in requests from bots where the usage of headers might be more or less static and blocking them by the fingerprint is a different use case from Anubis, which generates a fingerprint for every client and, importantly, requires the fingerprint to be more or less constant for the client, since additional user interaction is warranted if the fingerprint changes.

docs/docs/developer/thr1.mdx
```diff
@@ -0,0 +1,187 @@
...
+  - The first 4 alphanumeric characters of the header value (lowercased, right-padded with `0` to length 4), e.g. `enca`.
+  - The first 9 hex characters of the SHA-256 hash of the full `Accept-Language` header value.
+
+Example:
+
+```
+enca-d6b272e5b
+```
+
+### `thr1_sec`
```
ryanccn, line 61: `Sec-CH-*` headers, unlike `Cookie` and `Referer`, are a great choice for fingerprinting because they actually identify interesting features about the client that might help uniquely identify them. However, `Sec-Fetch-*` headers are dependent on how the user interacts with the user agent and the website, so I don't think these headers are suitable for inclusion in a fingerprint. For example, requests for JavaScript modules or images have vastly different `Sec-Fetch-*` headers to direct navigation requests by the user, so these requests, from the same client, would end up having different fingerprints.

docs/docs/developer/thr1.mdx
```diff
@@ -0,0 +1,187 @@
...
+platform:windows
+platform_version:10.0.0
+sec-fetch-dest:document
+sec-fetch-mode:navigate
+ua:Chromium/123,Google Chrome/123
+```
+
+### `thr1_all`
+
+A hash of the canonicalized form of request headers.
```
ryanccn: This is a big problem, in my opinion, because of the same reason as many of the previous comments: a lot of headers are absolutely not static for the same client. For example, `Content-Type` is set when submitting multipart forms. Should the user suddenly have a different fingerprint because they are submitting a form? I don't think so.

ryanccn:
In summary, and I am sorry to say this, but it feels like the majority of this spec does not have a material improvement over the existing implementation, other than the use of the HTTP version and the `Sec-CH-*` headers. All of the other additions, including the request method, the presence of cookies/referrer, the `Sec-Fetch-*` headers, and all the rest of the headers, add more information to the fingerprint but also make it so that a human user could have drastically different fingerprints during their normal usage of a website.

As an experiment, I implemented the THR1 algorithm in JavaScript and tested it by logging fingerprints of requests to a local development server. Upon opening a very simple page, only five HTTP requests were made and there were four different THR1 fingerprints between them.

docs/docs/developer/thr1.mdx
felix9: JA4H is intended to identify bots, not humans. bots tend to have consistent behavior and will have similar or identical fingerprints. the goal of JA4 signatures is not "track a single user's session", it's "identify the same actor that's using multiple IP addresses"

docs/docs/developer/thr1.mdx
```diff
@@ -0,0 +1,187 @@
...
+   ```
+
+   Joined by newlines.
+
+4. Compute the SHA-256 checksum of that string and take the first 9 hex digits.
+
+Example output:
+
+```text
+2a347fcf7
```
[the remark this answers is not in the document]
janbrennen: bruh chill

docs/docs/developer/thr1.mdx
Xe: Ironically, a referer being present is a more realistic indicator that they're a human.

Issue #525, opened 2025-05-20
Reuse X-Forwarded-For header


derritter88:
Hello there,

my current setup network-wise would be:

WAN -> Nginx on Gitlab server which does TLS termination (sees real IP) -> Anubis deployed at Kubernetes and gets X-Forwarded-For Header -> Gitlabs own Nginx server on a different port which sees only the ClusterIP address from Anubis.

Nginx with TLS termination access.log:
```
10.0.22.70 - - [20/May/2025:10:42:52 +0200] "POST /api/v4/jobs/request HTTP/1.1" 204 0 "-" "gitlab-runner 17.9.1 (17-9-stable; go1.23.2 X:cacheprog; linux/amd64)"
```

Gitlabs Nginx after Anubis access.log:
```
10.42.3.0 - - [20/May/2025:10:43:10 +0200] "POST /api/v4/jobs/request HTTP/1.1" 204 0 "" "gitlab-runner 17.9.1 (17-9-stable; go1.23.2 X:cacheprog; linux/amd64)" -
```

derritter88:
@Xe 
Hello there,

any idea how to keep the origin IP address forwarded until the reverse proxied services?

derritter88:
For completness my current Nginx forwarding configuration would be:

```
        location / {
                if ($blocked_host) {
                        return 403;
                }

                proxy_set_header Host $host;
                proxy_set_header X-Real-IP $remote_addr;
                proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
                proxy_set_header X-Forwarded-Proto $scheme;
                proxy_pass http://10.0.22.118:8081;
                proxy_http_version 1.1;
                proxy_set_header   Upgrade $http_upgrade;
                proxy_set_header   Connection "upgrade";
        }
```

dchandekstark:
@derritter88 We need to be able to toggle off the `StripPrivate` setting at https://github.com/dchandekstark/anubis/blob/main/internal/headers.go#L89.

dchandekstark:
See also https://github.com/TecharoHQ/anubis/issues/455

derritter88:
@dchandekstark I agree - In general I would always use private IPs as at least in my environment everything is being reveresed proxyed and using a private IP.

ciencia:
My "solution" was to add a fake "X-Forwarded-For-Frontend" header from the frontend nginx:

```
	proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
[...]

Issue #455, opened 2025-05-06
x-forwarded-for private addresses are skipped, and there is a parsing bug


dchandekstark:
I had a very simple setup in k8s with anubis behind an OKD route.  Everything worked perfectly except that USE_REMOTE_ADDRESS didn't do what I want, since the client's address is in x-forwarded-for.

dchandekstark:
Update - I think this is arguably a bug? Private addresses are automatically skipped in x-forwarded-for, so if you're on a VPN, for example, your address gets dropped and x-real-ip doesn't get set unless you turn on USE_REMOTE_ADDRESS.

So, I'm not sure the request as originally stated is correct, since anubis does try to use x-forwarded-for.

dchandekstark:
https://github.com/TecharoHQ/anubis/blob/main/internal/headers.go#L89 - allow the StripPrivate option to be toggled off?

nate-duke:
From my naive reading of the go code in the headers handling if you are behind a proxy that's sending through X-Forwarded-For with a private IP address (i.e. vpn, local network, etc) the StripPrivate preference passed to [computeXFFHeader](https://github.com/TecharoHQ/anubis/blob/2e9b18a510821bd1bf7390b4147109171c9ca0fd/internal/headers.go#L124C6-L124C22) will cause it to get dropped, which clears X-Real-Ip later in [XForwardedForToXRealIP](https://github.com/TecharoHQ/anubis/blob/2e9b18a510821bd1bf7390b4147109171c9ca0fd/internal/headers.go#L70C6-L70C28)

Again, my naive TLDR; private IPs entering anubis through a proxy that sends X-Forwarded-For results in empty X-Real-Ip and the inability to skip challenges on local requests.

dchandekstark:
I think there's also a bug at https://github.com/TecharoHQ/anubis/blob/main/internal/headers.go#L136 which leads to errors like:

{"time":"2025-05-06T14:31:54.568190958Z","level":"DEBUG","source":{"function":"github.com/TecharoHQ/anubis/internal.computeXFFHeader","file":"github.com/TecharoHQ/anubis/internal/headers.go","line":158},"msg":"failed to parse XFF segment","err":"ParseAddr(\" 10.128.114.86\"): unexpected character (at \" 10.128.114.86\")"}

Need to trim spaces.

dchandekstark:
[...]

PR #619, opened 2025-06-05
Adds ability to toggle off stripping of private addrs from XFF

dchandekstark:
This is a conversation starter about how to make it possible to toggle off stripping of private addresses from XFF.  There is a comment in the code about making this available in config, so the method used here is presumably not the desired one (and not the most efficient).  I'm kind of hazy on the cli/config/environment details.

See also #455 and #525.

Checklist:


internal/headers.go
```diff
@@ -206,3 +212,16 @@ func NoBrowsing(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+// lookupEnvBool returns the boolean value of an environment variable if set and parsable, otherwise false and ok=false.
+func lookupEnvBool(key string) (bool, bool) {
```
dchandekstark: This function should panic or something on error.

Xe: TBH the second return is superfluous

dchandekstark: Feel free to mod to your style.  I'm only a some-time Golanger. ;)

PR #647, opened 2025-06-11
refactor(ogtags): optimize URL construction and memory allocations

JasonLovesDoggo:
reduced memory usage in GetTarget, cutting allocations by 2–8 bytes per op (up to 20% on Unix). ExtractOGTags memory stayed the same, but performance improved across the board — up to 1.66× faster in some cases.

Checklist:


JasonLovesDoggo:
not inlining `u.EscapedPath()` reduces runtime from `70.89 ns/op` to 	`48.77 ns/op`

internal/ogtags/ogtags.go
```diff
@@ -13,8 +13,9 @@ import (
 )
 
 const (
-	maxContentLength = 16 << 20        // 16 MiB in bytes, if there is a reasonable reason that you need more than this...Why?
-	httpTimeout      = 5 * time.Second /*todo: make this configurable?*/
+	maxContentLength = 8 << 20 // 8 MiB in bytes,
+	// if there is a reasonable reason that you need more than this...Why?
```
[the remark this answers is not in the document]
Xe:
```suggestion
	maxContentLength = 8 << 20 // 8 MiB is enough for anyone
```

Xe (approved):
We probably need some kind of ram limit for the ogtags cache, but this would work yeah!

JasonLovesDoggo:
> We probably need some kind of ram limit for the ogtags cache, but this would work yeah!

1000% I think that's more of a decaymap feature but this is a great start

JasonLovesDoggo:
I added the fuzz tests just due to my custom parsing. Sorta worried about a case I didn't consider but-

Xe (approved):

internal/ogtags/ogtags.go
```diff
@@ -71,37 +76,50 @@ func NewOGTagCache(target string, ogPassthrough bool, ogTimeToLive time.Duration
...
 
-	// For regular http/https targets
-	target := *c.targetURL // Make a copy
-	target.Path = u.Path
-	target.RawQuery = u.RawQuery
-	return target.String()
+	// For regular http/https targets, build URL string directly
+	var sb strings.Builder
+	// Pre-calculate size: scheme + "://" + host + path + "?" + query
+	estimatedSize := len(c.targetURL.Scheme) + schemeSeparatorLength + len(c.targetURL.Host) + len(escapedPath) + len(u.RawQuery) + querySeparatorLength
```
guest20, line 110: Doesn't `net/url` [do the same thing?](https://cs.opensource.google/go/go/+/refs/tags/go1.24.4:src/net/url/url.go;drc=bb5339196f86057ffa6042d31f52a44bf7f13752;l=846)

JasonLovesDoggo: I can't remember my benchmarks on it and I'll try it again but from my memory that was two allocations and about 1.66 times slower. 

I'll try it again and update you

PR #680, opened 2025-06-17
Add option to use HS512 secret for JWT instead of ED25519

Earl0fPudding:
Due to the fact that unfortunately a lot of software isn't yet compatible with ED25519 JWT, it would be nice if one has the option to use HS512 instead of ED25519.

In particular, this can be used in conjunction with HAProxy as it has the ability to check the JWT payload and verify it directly. One can then configure that requests with the correct JWT will go directly to the desired backends, whereas unverified requests firstly go through Anubis.
Example Config snippet for HAProxy:
```
  # use anubis is cookie not found
  use_backend BE-lighttp-via-anubis unless { req.cook(techaro.lol-anubis-auth) -m found }

  # get data from JWT
  http-request set-var(txn.alg) req.cook(techaro.lol-anubis-auth),jwt_header_query('$.alg')
  http-request set-var(txn.iss) req.cook(techaro.lol-anubis-auth),jwt_payload_query('$.iss')
  http-request set-var(txn.aud) req.cook(techaro.lol-anubis-auth),jwt_payload_query('$.jti')
  http-request set-var(txn.aud) req.cook(techaro.lol-anubis-auth),jwt_payload_query('$.exp')

  # use anubis if wrong algo
  use_backend BE-lighttp-via-anubis unless { var(txn.alg) -m str HS512 }

  # use anbis if expired
  http-request set-var(txn.now) date()
  use_backend BE-lighttp-via-anubis if { var(txn.exp),sub(txn.now) -m int lt 0 }

  # use anubis if wrong signature
  use_backend BE-lighttp-via-anubis  unless { req.cook(techaro.lol-anubis-auth),jwt_verify(txn.alg,"SECRET") -m int 1 }

  default_backend BE-lighttp-direct
```
This would mean that only the first request has to go through Anubis.
Unfortunately HAProxy is not compatible with ED25519, HS512 should suffice for this case.

This is not a breaking change of Anubis. If nothing is changed in the config, then ED25519 will be used. Only if the new `HS512_SECRET` config value is being set explicitly, HS512 will be used.

Please let me know if I should adjust something, thanks.
Perhaps someone could help me with the testcases if needed.

Many thanks, 
Martin

Checklist:

- [ v] Added a description of the changes to the `[Unreleased]` section of docs/docs/CHANGELOG.md
- [ x] Added test cases to [the relevant parts of the codebase](https://anubis.techaro.lol/docs/developer/code-quality)
- [ v] Ran integration tests `npm run test:integration` (unsupported on Windows, please use WSL)


cmd/anubis/main.go
```diff
@@ -48,6 +48,7 @@ var (
 	cookieDomain             = flag.String("cookie-domain", "", "if set, the top-level domain that the Anubis cookie will be valid for")
 	cookieExpiration         = flag.Duration("cookie-expiration-time", anubis.CookieDefaultExpirationTime, "The amount of time the authorization cookie is valid for")
 	cookiePartitioned        = flag.Bool("cookie-partitioned", false, "if true, sets the partitioned flag on Anubis cookies, enabling CHIPS support")
+	hs512Secret              = flag.String("hs512-secret", "", "secret used to sign JWTs, uses ed25519 if not set")
```
Xe, line 51: How is one of these secrets generated?

cmd/anubis/main.go
```diff
@@ -290,11 +291,15 @@ func main() {
...
 	}
 
-	var priv ed25519.PrivateKey
-	if *ed25519PrivateKeyHex != "" && *ed25519PrivateKeyHexFile != "" {
+	var ed25519Priv ed25519.PrivateKey
+	if *hs512Secret != "" && (*ed25519PrivateKeyHex != "" || *ed25519PrivateKeyHexFile != "") {
+		log.Fatal("do not specify both HS512 and ED25519 secrets")
+	} else if *hs512Secret != "" {
+		ed25519Priv = ed25519.PrivateKey(*hs512Secret)
+	} else if *ed25519PrivateKeyHex != "" && *ed25519PrivateKeyHexFile != "" {
```
Xe, line 299: Probably could become a switch case block as that's more idiomatic Go

docs/docs/admin/installation.mdx
```diff
@@ -58,6 +58,7 @@ Anubis uses these environment variables for configuration:
...
 | `ED25519_PRIVATE_KEY_HEX_FILE` | unset                   | Path to a file containing the hex-encoded ed25519 private key. Only one of this or its sister option may be set.                                                                                                                                                                                                                                                                                                                                                                                        |
+| `HS512_SECRET`                 | unset                   | Secret string for JWT HS512 algorithm. If this is not set, Anubis will use ED25519 as defined via the variables above.                                                                                                                                                                                                                                                                                                                                                                                  |
```
Xe: Move this to the advanced options section please.

lib/anubis.go
```diff
@@ -154,9 +159,16 @@ func (s *Server) maybeReverseProxy(w http.ResponseWriter, r *http.Request, httpS
...
+	var token *jwt.Token
+	if len(s.hs512Secret) == 0 {
+		token, err = jwt.ParseWithClaims(ckie.Value, jwt.MapClaims{}, func(token *jwt.Token) (interface{}, error) {
+			return s.ed25519Priv.Public().(ed25519.PublicKey), nil
+		}, jwt.WithExpirationRequired(), jwt.WithStrictDecoding())
+	} else {
+		token, err = jwt.ParseWithClaims(ckie.Value, jwt.MapClaims{}, func(token *jwt.Token) (interface{}, error) {
+			return s.hs512Secret, nil
+		}, jwt.WithExpirationRequired(), jwt.WithStrictDecoding())
+	}
```
Xe: I feel like there should be a better way to do this. Maybe a function on the Server type that returns the public key value as an any?

cmd/anubis/main.go
Earl0fPudding: Any string can be used as secret for HS512, the string will be converted as byte array. The longer the string the better, but it will take a tiny bit longer. 128 byte should suffice.
I can add more information if you wish.

cmd/anubis/main.go
Earl0fPudding: Yeah, that's true and I also thought of something like that. I just adapted it to fit the existing code :sweat_smile:

docs/docs/admin/installation.mdx
Earl0fPudding: Okay, will do.

lib/anubis.go
Earl0fPudding: The problem is that it throws an error if i just replace the value as for HS512, the function expects a byte array.
I could implement a function on the Server type which returns the whole token `keytype`.

docs/docs/admin/installation.mdx
Earl0fPudding: Moved.

lib/anubis.go
Earl0fPudding: Done. The function `getTokenKeyfunc` has been introduced for Server type. 
Now the code here has been reduced to:
```
token, err := jwt.ParseWithClaims(ckie.Value, jwt.MapClaims{}, s.getTokenKeyfunc(), jwt.WithExpirationRequired(), jwt.WithStrictDecoding())
```

Earl0fPudding:
Should I re-do this branch and pull request with signed commits or can you accept it without them?

Earl0fPudding:
```
### Unrecognized words (1)
Keyfunc
```
Should we "allowlist" the word? `Keyfunc` is a type in the JWT library.

EDIT: Done.

PR #682, opened 2025-06-17
feat(config): Add FCrDNS checker

### LINKED ISSUE #431: Dynamic validation of good bot IP addresses
AFAICT Anubis currently uses hard-coded lists of IP addresses to detect impostors of good bots, which might fall behind the current IPs those services are using if Anubis's update cadence doesn't line up. Bing in particular specifically advises against hard-coding their IP ranges as they are subject to change without any warning.

https://www.bing.com/webmasters/help/how-to-verify-bingbot-3905dc26

> **Don’t Use Hardcoded IP Addresses or Address Ranges**
>
> Use the reverse/forward DNS lookup method you can easily verify that an IP address is coming from Bing.
> 
> Or alternatively, you can identify Bingbot by IP address by matching the crawler's IP address to the [list of Bingbot IP addresses](https://www.bing.com/toolbox/bingbot.json). **Refresh this list daily as we can change this list any time.**

Would it be worth implementing DNS-based validation for Googlebot and Bingbot?

Axelen123:
Closes #431. This PR implements dynamic verification of bot IPs using DNS records. For details regarding how it works, see the documentation I have added. I am not sure if the way I added a new `algorithm` is the best way to implement this. Let me know if there is a better way.


Xe:
Hey, thanks for the contribution!

The CHALLENGE rule is more meant for client-facing challenges. You probably want something like the [checker.Impl](https://github.com/TecharoHQ/anubis/blob/main/lib/policy/checker/checker.go#L12-L15) interface here. This will let you add an `ALLOW` rule for AppleBot et.al.

lib/policy/checker/checker.go
```diff
@@ -19,15 +19,12 @@ type List []Impl
...
+		if err != nil || !ok {
 			return ok, err
 		}
-		if ok {
-			return ok, nil
-		}
 	}
 
-	return false, nil
+	return true, nil
```
Axelen123: This change was made to ensure that we don't try to do any unnecessary reverse DNS checks if other checkers (such as user agent or CEL) fail. I assumed this was already the case and I am not sure if changing this will have unintended consequences.

Axelen123:
I have finished converting the implementation to a checker.

lib/policy/checker/checker.go
Xe: May be best to change it back, false means "no match", true means "match".

Xe (approved):
Approved modulo the change to checker.List#Check

lib/policy/checker/checker.go
Axelen123: Still seems weird to me because I, as a user, would expect that in order for a rule to match, the request has to match all conditions on it. I can change it back, but I will mark this PR as draft while I add CEL support for this feature. Otherwise, each default crawler rule with `domain_regex` would result in a DNS request, even for clients that do not match the user agent regex.

Axelen123:
I have added CEL bindings and reverted to the previous checker behavior.

Xe (approved):

Xe:
Thanks much! This is gonna let us do a lot of fun things :)

Axelen123:
I was reminded of this PR recently. Did you forget to merge it or is there something else?
Let me know if there is anything else I need to do (and thank you for the other commits)

Xe:
I forgot about this, I'm going to rebase after work.

PR #685, opened 2025-06-17
feat(lib): Add option for automatically setting cookie domain as second level domain

Earl0fPudding:
Normally, one can only choose between setting the cookie domain as the exact domain of the host header (`COOKIE_DOMAIN` is empty) or as a static string (`COOKIE_DOMAIN` contains domain).

If someone has multiple domains and subdomains (e.g. www.domain.tld, subdomain.domain.tld, www.domain2.tld), then the users have to solve the PoW for every subdomain (`COOKIE_DOMAIN` is empty) or only allow one of the domains as a whole (`COOKIE_DOMAIN=domain.tld`) without multiple Anubis instances.

With those changes, one can set `COOKIE_DOMAIN` to `DYNAMIC_SECOND_LEVEL_DOMAIN` and Anubis automatically parses the second level domain (`subdomain.domain.tld` -> `domain.tld`) from the host header and sets it as cookie domain. Thus multiple domains and subdomains can be protected but the PoW is only needed once per whole second level domain, no multiple instances needed. 
The current behaviour regarding this variable is not being altered, this is not a breaking change.

If testcases are needed, perhaps someone can help me with that.

Many thanks,
Martin

Checklist:


Xe (approved):

Xe:
When I was doing pre-merge testing to ensure everything was all good, I ran into strange issues. I was unable to debug them so I tried recreating the core of the change in #722.

PR #694, opened 2025-06-18
feat: add default OpenGraph tags to configuration file

### LINKED ISSUE #681: Option to return default OpenGraph tags instead of looking them up with the remote target
A request from a customer. This should be done in the policy file, something like:

```yaml
opengraph:
  enable: true
  default: # map[string]string
```

Xe:
Closes #681

Checklist:

- [x] Documentation for this, maybe rename the policy file to the configuration file?


internal/ogtags/cache_test.go
```diff
@@ -7,10 +7,49 @@ import (
...
 	"testing"
 	"time"
+
+	"github.com/TecharoHQ/anubis/lib/policy/config"
 )
 
+func TestCacheReturnsDefault(t *testing.T) {
+	want := map[string]string{
+		"og:title":       "Foo bar",
+		"og:description": "The best website ever made!!!1!",
```
JasonLovesDoggo, line 17: based test data

internal/ogtags/cache.go
```diff
@@ -13,6 +13,10 @@ func (c *OGTagCache) GetOGTags(url *url.URL, originalHost string) (map[string]st
 		return nil, errors.New("nil URL provided, cannot fetch OG tags")
 	}
 
+	if len(c.ogDefault) != 0 {
```
JasonLovesDoggo: The downside is this limits per-route default syntax but that's out of scope here

internal/ogtags/ogtags_test.go
```diff
@@ -175,7 +189,9 @@ func TestGetTarget(t *testing.T) {
 func TestIntegrationGetOGTags_UnixSocket(t *testing.T) {
 	tempDir := t.TempDir()
 
-	socketPath := filepath.Join(tempDir, "anubis-test.sock")
+	// XXX(Xe): if this is named longer, macOS fails with `bind: invalid argument`
+	// because the unix socket path is too long. I love computers.
+	socketPath := filepath.Join(tempDir, "t")
```
JasonLovesDoggo, line 194: i- wow okay. I've always hated file name too long bugs

lib/policy/config/config.go
```diff
@@ -353,10 +361,21 @@ func Load(fin io.Reader, fname string) (*Config, error) {
...
+		DNSBL: c.DNSBL,
+		OpenGraph: OpenGraph{
+			Enabled:      c.OpenGraph.Enabled,
+			ConsiderHost: c.OpenGraph.ConsiderHost,
+			Default:      c.OpenGraph.Default,
+		},
 		StatusCodes: c.StatusCodes,
 	}
 
+	if c.OpenGraph.TimeToLive != "" {
```
JasonLovesDoggo, line 396: is there a default? How does the parser handle a opengraph block with no TTL

lib/policy/config/opengraph.go
```diff
@@ -0,0 +1,51 @@
...
+func (og *openGraphFileConfig) Valid() error {
+	var errs []error
+
+	if _, err := time.ParseDuration(og.TimeToLive); err != nil {
+		errs = append(errs, fmt.Errorf("%w: ParseDuration(%q) returned: %w", ErrOpenGraphTTLDoesNotParse, og.TimeToLive, err))
+	}
+
+	if len(og.Default) != 0 {
+		for _, tag := range []string{
+			"og:title",
```
JasonLovesDoggo, line 38: why require  og:title? Also i'd move that string slice out into something like requiredTags to be a little more clear

JasonLovesDoggo (approved):

JasonLovesDoggo:
Besides my comments, assuming it's tested, LGTM

lib/policy/config/opengraph.go
Xe: From what I've seen every service that parses open graph tags needs at least og:title.

JasonLovesDoggo:
one change i'd like is to rename "default" to some other keyword as it's not a fallback but a primary mechanism to get/return og tags

JasonLovesDoggo (approved):

PR #697, opened 2025-06-19
feat(lib): Add optional restrictions for JWT based on a specific header value

Earl0fPudding:
Due to the fact that it would be possible for a botnet to generate the JWT once and then roll it out on many bots, it would be nice if one can restrict the JWT to specific criteria.
With this change, one can define the `JWT_RESTRICTION_HEADER` setting which allows the token to only be valid if a specific header value is the same as it was during the creation of the token. Setting it to `X-Real-IP` e.g. would restrict the token only to be valid for this one IP address.

One can also add custom header using a reverse-proxy using a combined values of SRC-IP, User-Agent, TLS-Fingerprint etc to restrict the token even further. E.g. in HAProxy:
```
  http-request add-header X-Custom-Anubis-Restriction "%[src]__%[req.fhdr(User-Agent)]"
```
And then setting `JWT_RESTRICTION_HEADER=X-Custom-Anubis-Restriction` in Anubis.

This may result in having to re-calculate the token when changing SRC-IPs for the user, but it might be necessary when dealing with an intelligent botnet.

SHA256 is being used as potentially sensitive information might be used for restrictions.

This is not a breaking change. If the setting is empty, then the behavior is as normal.

Checklist:


Xe:
I think this should be on by default and set to X-Real-Ip.

Earl0fPudding:
> I think this should be on by default and set to X-Real-Ip.

Thanks, what an honor you think this should be set by default. `X-Real-IP` as default has been set.

JasonLovesDoggo:
> I think this should be on by default and set to X-Real-Ip.

Is that not already done? https://github.com/TecharoHQ/anubis/blob/226cf36bf7707625dd756fb63f4164fefbbf10e6/lib/anubis.go#L86

Earl0fPudding:
> > I think this should be on by default and set to X-Real-Ip.
> 
> Is that not already done?

Do you mean something like that in the current main branch, @JasonLovesDoggo ? Not really unfortunately. 
I just tested this myself. I compiled the latest main, rolled it out on my server and called the website from the same device and same browser but from different IPs. The result was that only the first time with the first IP the token has to be generated, the other times with other IPs Anubis let me straight to the website.

This means that currently it is not possible to restrict the token to a specific attribute (e.g. SRC IP) and botnets could use this to their advantage. Nothing is easier than calling the website manually one and roll a cookie out on thousands of their bots. With the default vault of `COOKIE_EXPIRATION_TIME`, this only needs to be done once a week.

PR #699, opened 2025-06-20
Makefile: Build robots2policy

heftig:
Make the `Makefile` build the `robots2policy` tool as well.

Adjust the documentation to mention this fact.


JasonLovesDoggo (approved):
LGTM, but just have to wait for Xe if this is how it should be packaged

heftig:
> LGTM, but just have to wait for Xe if this is how it should be packaged

For context, see https://gitlab.archlinux.org/archlinux/packaging/packages/anubis/-/blob/main/PKGBUILD

PR #716, opened 2025-06-24
feat: implement localization system

### LINKED ISSUE #449: i18n: Translations of the Anubis UI
This is a big one. Holy wow this is a big one.

The hard part with this is that there's technically two programs to be localized here:

* The backend Anubis code
* The frontend Anubis code

Any solution we make will inevitably be better for one over the other. I think that it may be best to prioritize the backend solution and then make the backend "embed" the strings the frontend may need in a JSON block like we do for the version and challenge data.

lolgzs:
Closes #205 
Closes #449 

Locale files are placed in lib/localization/locales/. If you add a locale, update manifest.json with available locales.

Note that I'm a go noob. Extensive use of AI to implement it ....

Checklist:

- [X] Added test cases to lib/localization/
- [ ] Ran integration tests `npm run test:integration`  : seems it needs external services ?

### DIFF
--- a/lib/localization/localization_test.go
+++ b/lib/localization/localization_test.go
@@ -58,6 +58,27 @@ func TestLocalizationService(t *testing.T) {
 	})
 }
 
+type manifest struct {
+	SupportedLanguages []string `json:"supported_languages"`
+}
+
+func loadManifest(t *testing.T) manifest {
+	t.Helper()
+
+	fin, err := localeFS.Open("locales/manifest.json")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer fin.Close()
+
+	var result manifest
+	if err := json.NewDecoder(fin).Decode(&result); err != nil {
+		t.Fatal(err)
+	}
+
+	return result
+}
+
 func TestComprehensiveTranslations(t *testing.T) {
 	service := NewLocalizationService()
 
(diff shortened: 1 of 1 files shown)


lib/anubis.go
```diff
@@ -343,34 +356,34 @@ func (s *Server) PassChallenge(w http.ResponseWriter, r *http.Request) {
...
-		s.respondWithError(w, r, "Redirect domain not allowed")
+		s.respondWithError(w, r, localizer.T("redirect_domain_not_allowed"))
 		return
 	}
 
 	cr, rule, err := s.check(r)
 	if err != nil {
 		lg.Error("check failed", "err", err)
-		s.respondWithError(w, r, "Internal Server Error: administrator has misconfigured Anubis. Please contact the administrator and ask them to look for the logs around \"passChallenge\"")
+		s.respondWithError(w, r, fmt.Sprintf("%s \"passChallenge\"", localizer.T("internal_server_error")))
```
avioletheart, line 378: the `\"passChallenge\"` should probably be inside the translation files, or passed as a parameter. not all languages order sentences in such a way where the object is at the end; for instance, a simplified japanese sentence order would be `[to the administrator] [logs around "passChallenge"] [look for] [ask]`.

PR #747, opened 2025-07-02
feat(web): Add option for customizable explanation text

Earl0fPudding:
Hi Xe,

I know, at this point I must be really annoying.
However, this PR adds the ability to replace the default explanation text when clicking "Why am I seeing this?" with a custom one.

This can be used because unfortunately other people seem to need a bit more dumbed down version of the explanation. Using this, it can be customized for your audience via the new `customExplanationText` option.

E.g. this in the config:
```
CUSTOM_EXPLANATION_TEXT="This is a Test."
```
results in the following site:
Screenshot from 2025-07-02 17-17-07

I'm not sure if test cases are necessary in this case. If yes, I'd be glad for your support.
Please let me know if there are any questions, thanks.

Long live the emperor,
Martin

Checklist:


Xe:
What is an example of this dumbed-down text? I think it may be better to have that as an option in the i18n subsystem.

Earl0fPudding:
> What is an example of this dumbed-down text? I think it may be better to have that as an option in the i18n subsystem.

Sorry for my absence, DebConf is consuming my time currently. I've rewritten this branch to feature a static simplified text instead of a custom one. Here is the text exatcly:
https://github.com/TecharoHQ/anubis/pull/747/files#diff-41754c99306958d5693233880f198ceae2f424f2aa8aa600a5d713cbe3efa316R10

What do you think, Xe?

PR #763, opened 2025-07-06
docs: remove proof of work branding

Xe:
- Redesign the homepage to remove proof of work from the branding as it has been causing confusion
- Use a custom social card hastily made with Figma
- Refine the toolbar links


docs/docusaurus.config.ts
```diff
@@ -6,7 +6,7 @@ import type * as Preset from '@docusaurus/preset-classic';
 
 const config: Config = {
   title: 'Anubis',
-  tagline: 'Weigh the soul of incoming HTTP requests using proof-of-work to stop AI crawlers',
+  tagline: 'Weigh the soul of incoming HTTP requests to protect your website!',
```
guest20, line 9: I think most of the users who see this message will not be the site owner...

PR #781, opened 2025-07-08
Update manifest.json

sveinki:
Adding Icelandic (is) as an option in manifest.json

Just trying to speed up things ;-)


JasonLovesDoggo:
Please put this in #780. Both PRs are useless without eachother and the changes are small

PR #784, opened 2025-07-08
Update apache.mdx

jzb:
Was missing the opening stanza to enable mod_proxy for Apache. Added that to the doc.


jzb:
This was on Debian bookworm.

PR #834, opened 2025-07-15
feat(cmd/anubis): capture ja4h fingerprints

Xe:
This is not used yet, but it will be part of a larger strategy around adding/removing weight based on JA4H (and other) fingerprint matches with Thoth.

Checklist:


jvoisin:
I was benchmarking anubis, and it seems that ja4h is on the hot path:

```go
package internal

import (
	"net/http"
	"net/http/httptest"
	"testing"
)

func realisticRequest() *http.Request {
	r := httptest.NewRequest(http.MethodGet, "/some/path?foo=bar", nil)
	r.Header.Set("User-Agent", "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0 Safari/537.36")
	r.Header.Set("Accept", "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8")
	r.Header.Set("Accept-Language", "en-US,en;q=0.9")
	r.Header.Set("Accept-Encoding", "gzip, deflate, br")
	r.Header.Set("Referer", "https://example.com/")
	r.Header.Set("Priority", "u=0, i")
	r.Header.Set("X-Forwarded-For", "203.0.113.10")
	r.Header.Set("X-Real-Ip", "203.0.113.10")
	r.AddCookie(&http.Cookie{Name: "within.website-x-cmd-anubis-auth", Value: "deadbeef"})
	r.AddCookie(&http.Cookie{Name: "session", Value: "0123456789abcdef"})
	return r
}

func BenchmarkRequestChainWithoutJA4H(b *testing.B) {
	h := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {})

	b.ReportAllocs()
	b.ResetTimer()
	for range b.N {
		r := realisticRequest()
		w := httptest.NewRecorder()
		h.ServeHTTP(w, r)
	}
}

func BenchmarkRequestChainWithJA4H(b *testing.B) {
	h := JA4H(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {}))

	b.ReportAllocs()
	b.ResetTimer()
	for range b.N {
		r := realisticRequest()
		w := httptest.NewRecorder()
		h.ServeHTTP(w, r)
	}
}
```

| Metric | With JA4H | Without JA4H | Improvement |
|--------|-----------|--------------|-------------|
|sec/op	 |25.7µs ± 28%|	8.52µs ± 22%|	−66.9% (p=0.000)|
|B/op	 |10,844	|6,614	|−39.0% (p=0.000)|
|allocs/op|	89|	30|	−66.3% (p=0.000)|

Since it's currently unused, and has been so for more than a year, would it make sense to put it behind a flag? Something like `if policy.NeedJA4H { h = internal.JA4H(h) }` in cmd/anubis/main.go? Or is its impact deemed acceptable?

Xe:
It's used by people's configs, I have a policy that people's configs are part of the public API. It may be better to just optimize the ja4h paths.

jvoisin:
The lazy way would be to have a `func configReferencesJA4H(bots []config.BotConfig) bool ` in `lib/policy/policy.go`, and use it in `cmd/anubis/main.go` for the `policy.NeedJA4H { h = internal.JA4H(h) }` condition, but it's a tad ugly.

Xe:
The lazier way would be to only upsert the JA4H header when it's looked up in [the header CEL bridge](https://github.com/TecharoHQ/anubis/blob/main/lib/policy/expressions/http_headers.go#L43). It'd be super hacky, but I see no reason it should fail.

jvoisin:
I'm not comfortable sending super-hacky pull-requests your way :P

PR #860, opened 2025-07-18
feat: Osiris, a TLS reverse proxy with fingerprinting powers Anubis can work with

Xe:
Checklist:


DemiMarie:
Curiosity question: why make this separate from Anubis proper?

Earl0fPudding:
@DemiMarie the reason was because this repo had already most dependencies and helper functions required for Osiris and it was just "quick and dirty" implemented here because of this. If Osiris is going to be developed further, I'm sure it will get its own repo.

DemiMarie:
@Earl0fPudding That makes sense.  I was actually expecting that Osiris would be part of the main Anubis process.

Xe:
@DemiMarie it likely will be in the future, but for now it's a separate binary to reduce dependency bloat.

PR #874, opened 2025-07-21
Revert "build(deps): bump the gomod group with 6 updates"

JasonLovesDoggo:
Reverts TecharoHQ/anubis#873


JasonLovesDoggo:
Builds worked before... not sure why they failed once merged in but I can't dig in right now

PR #878, opened 2025-07-21
Update pt-BR.json

HQuest:
While current version is good enough as a machine translation, it is not natural enough for what a reader would like to read while browsing sites made by native devs - including the subtle nuances from the original English version, now incorporated to the translation instead of plain, literal translations with questionable meanings.


lib/localization/locales/pt-BR.json
```diff
@@ -2,19 +2,19 @@
   "loading": "Carregando...",
   "why_am_i_seeing": "Por que estou vendo isso?",
   "protected_by": "Protegido por",
-  "protected_from": "From",
-  "made_with": "Feito com ❤️ no 🇨🇦",
+  "protected_from": "de",
+  "made_with": "Feito com ❤️ na Califórnia",
```
victorvalenca:
```suggestion
  "made_with": "Feito com ❤️ no Canadá",
```

victorvalenca (commented):
Looks good! Just one change needed here that might've been from the fork this is coming from

PR #880, opened 2025-07-21
fix(web/fast): remove event loop thrashing after 1024 iterations

### LINKED ISSUE #877: Challenge: JS memory exhaustion on 32-bit browsers
The current proof-of-work "temporary" solution seems to cause memory exhaustion on 32-bit browsers, as reported by our (Pale Moon) users.
Especially in situations where the Anubis difficulty is cranked up (e.g. Devuan website resources where it's set to `5`), the situation causes browser crashes due to exhaustion of the JS stack on 32-bit builds -- this is most likely because the algorithm's workers swamp the event loop and don't allow proper garbage collection in single-threaded JS engines. The behaviour is primarily observed if users continue to either interact with the browser while the challenge is running, or if the JS interrupt frequency is lowered by switching tabs or alt-tabbing to other applications. When JS interrupts get fired for non-anubis processing is where we run into OOM situations (stack space by the looks of it) that triggers safe crashes (deliberate).

Several people in our community have already tried to analyse this but it seems to be an inherent result of the Anubis algo causing too much call recursion, memory fragmentation and simply loading up the engine with tasks with no regard that it may need a few breathers for its internal administration. Long-running workers that cause 100% CPU load are problematic. It doesn't seem like we can do much to mitigate this on our end -- since the point of the PoW is to waste compute time anyway, is it possible to reduce the load caused by workers at the expense of hashing speed?

Once more: this seems to be entirely the case for 32-bit only, as I haven't been able to reproduce this on 64-bit browser builds. There doesn't seem to be much we can do in our architecture because we do not off-load tasks to separate browser processes. Users have indicated Firefox also suffered from this same issue on 32-bit up to and including ESR 115.

Please see the discussion on our forum for more detail: [topic 32549](https://forum.palemoon.org/viewtopic.php?f=70&t=32549)

Xe:
Fixes #877

Continued from #879, event loop thrashing can cause stack space exhaustion on ia32 systems. Previously this would thrash the event loop in Firefox and Firefox derived browsers such as Pale Moon. I suspect that this is the ultimate root cause of the bizarre irreproducible bugs that Pale Moon (and maybe Cromite) users have been reporting since at least #87 was merged.

The root cause is an invalid boolean statement:

```js
// send a progress update every 1024 iterations. since each thread checks
// separate values, one simple way to do this is by bit masking the
// nonce for multiples of 1024. unfortunately, if the number of threads
// is not prime, only some of the threads will be sending the status
// update and they will get behind the others. this is slightly more
// complicated but ensures an even distribution between threads.
if (
  (nonce > oldNonce) | 1023 && // we've wrapped past 1024
  (nonce >> 10) % threads === threadId // and it's our turn
) {
  postMessage(nonce);
}
```

The logic here looks fine but is subtly wrong as was reported in #877 by a user in the Pale Moon community. Consider the following scenario:

`nonce` is a counter that increments by the worker count every loop. This is intended to spread the load between CPU cores as such:

| Iteration | Worker ID | Nonce |
| :-------- | :-------- | :---- |
| 1         | 0         | 0     |
| 1         | 1         | 1     |
| 2         | 0         | 2     |
| 3         | 1         | 3     |

And so on.

The incorrect part of this is the boolean logic, specifically the part with the bitwise or `|`. I think the intent was to use a logical or (`||`), but this had the effect of making the `postMessage` handler fire on every iteration. The intent of this snippet (as the comment clearly indicates) is to make sure that the main event loop is only updated with the worker status every 1024 iterations per worker. This had the opposite effect, causing a lot of messages to be sent from workers to the parent JavaScript context.

This is bad for the event loop.

Instead, I have ripped out that statement and replaced it with a much simpler increment only counter that fires every 1024 iterations. Additionally, only the first thread communicates back to the parent process. This does mean that in theory the other workers could be ahead of the first thread (posting a message out of a worker has a nonzero cost), but in practice I don't think this will be as much of an issue as the current behaviour is.

The root cause of the stack exhaustion is likely the pressure caused by all of the postMessage futures piling up. Maybe the larger stack size in 64 bit environments is causing this to be fine there, maybe it's some combination of newer hardware in 64 bit systems making this not be as much of a problem due to it being able to handle events fast enough to keep up with the pressure.

Either way, thanks much to @wolfbeast and the Pale Moon community for finding this. This will make Anubis faster for everyone!

Checklist:


Xe:
Unscientific testing, but this provides an unbelievable speedup for Chrome and Firefox on my test hardware (MacBook Pro M3 Max):

| Browser | Before | After |
| :-- | :-- | :-- |
| Firefox | 48 kh/sec | 64 kh/sec |
| Google Chrome | 200-ish kh/sec | 250-ish kh/sec |

I'm still not sure why Chrome is so much faster. If I had to guess, Chrome is able to directly jump between JIT code and WebCrypto without having to do security overhead.

JasonLovesDoggo:
> I'm still not sure why Chrome is so much faster. If I had to guess, Chrome is able to directly jump between JIT code and WebCrypto without having to do security overhead.

image

Frosne:
> Unscientific testing, but this provides an unbelievable speedup for Chrome and Firefox on my test hardware (MacBook Pro M3 Max):
> Browser 	Before 	After
> Firefox 	48 kh/sec 	64 kh/sec
> Google Chrome 	200-ish kh/sec 	250-ish kh/sec
> 
> I'm still not sure why Chrome is so much faster. If I had to guess, Chrome is able to directly jump between JIT code and WebCrypto without having to do security overhead.

Hi!.
Could you share the code you used for testing, because I can not reproduce your results. 

Thanks!

PR #900, opened 2025-07-24
feat: Add thai language.

karorogunso:
Add Thai Language.


karorogunso:
Will keep this branch exist because there might be new text in the future.

PR #901, opened 2025-07-24
Pass forward X-Real-IP to nginx backend server

wolfbeast:
The TLS termination server sets X-Real-IP to be used by the back-end, but the back-end configuration example doesn't actually extract it so nginx logs (and back-end processing) fails to log or use the visiting IP in any way (it just states `unix:` if using a unix socket like in the example given, or the local IP if forwarded over TCP).

Adding real_ip_header to the config will fix this.

Checklist:

(this is just a doc change, the checklist didn't apply for this)


Xe (approved):
Thanks for the catch!

PR #926, opened 2025-07-27
feat(localization): Add Vietnamese translation

ninfia:
Adding Vietnamese language translation. Let me know if you have any question/feedback :)

Checklist:

- [X] Added test cases to [the relevant parts of the codebase](https://anubis.techaro.lol/docs/developer/code-quality)


lib/localization/locales/vi.json
```diff
@@ -0,0 +1,64 @@
+{
+  "loading": "Đang nạp...",
```
jaredallard, line 2: nạp is usually more charging a battery or adding

```suggestion
  "loading": "Đang tải...",
```

lib/localization/localization_test.go
```diff
@@ -26,6 +26,7 @@ func TestLocalizationService(t *testing.T) {
 		"pt-BR": "Carregando...",
 		"tr":    "Yükleniyor...",
 		"ru":    "Загрузка...",
+		"vi":    "Đang nạp...",
```
jaredallard, line 29:
```suggestion
		"vi":    "Đang tải...",
```

jaredallard (approved):
overall seems good, would be nice to get this in @Xe (when you can!)

lib/localization/locales/vi.json
ninfia: Charging the battery is "Đang sạc" (pin).

As for the decision of "nạp" instead of "tải", both technically means "load", but in the case "tải" it's more commonly used in upload/download in computing context, and "Đang tải/tải xuống" is a common short way to say "Downloading" (and for uploading it's "tải lên" since it's less common)

Hence the usage of "nạp" to distinct "Loading" instead of "Downloading".

jaredallard: ahhh, that makes sense, thanks for the explanation!

lib/localization/localization_test.go
jaredallard: apparently I can't resolve my own comment w/o write perms? but pls ignore!

Issue #949, opened 2025-08-03
[Feature] Add more favicon extensions by default (PNG, SVG)


OdinVex:
More browsers support SVG and PNG for favicons. Regarding `data/common/keep-internet-working.yaml`, could we extend the favicon to allow for png/svg? PNG support has been around for a very long time now across nearly every browser. I know I can add a policy, I just think it'd be interesting to show in that particular policy file a regex example for multiple extensions for such a common file.

https://caniuse.com/?search=favicon

arcayr:
something like the following should be what you're after:

```
path_regex: ^/favicon\.(?:ico|png|gif|jpg|jpeg|svg)$
```

* `(?: ... )` is a non-capturing group; the extension is irrelevant for the purposes of the regex engine - we only care that it exists. per my reading of go's `regexp` package,  it's slightly faster than a capturing group (no need to allocate / copy the group match).
* `favicon\.` matches "favicon.", with the "." appropriately escaped. previous rules in this file don't escape the '.', which technically means "faviconAico" or "favicon1ico" would match. i also need to file a pr to fix this.
* `ico|png|gif|jpg|jpeg|svg` is a collection of composites representing the relevant extensions. caniuse reports specifically on ico, png, and svg, however browsers typically support the full list of patterns here. it barely matters at all unless you're running anubis on a ti-89, but these are matched left-to-right, so if you're only using one or two of these extensions you can trim the list.

i filed a draft pr with the relevant changes [here](https://github.com/TecharoHQ/anubis/pull/961) if you're immediately interested given it's a configuration change that can be done on the current release.

technically speaking i guess you could alter it to the following:

```
path_regex: ^/favicon.*$
```

however `/favicon<anything>` would then match and bypass.

i tested the above against `HEAD`, not against the latest release, but the relevant file hasn't changed since that release.

OdinVex:
I was unaware of browsers supporting more than ico, png, and svg (forgot about gif), nice to know. :) I'm aware of the regex, I had already adjusted for `path_regex: ^/favicon\.(?:ico|png|svg)$` but I'll add the rest, thank you again. :)

PR #961, opened 2025-08-06
feat(checker): allow png/gif/jpg/jpeg/svg favicons as well as ico

arcayr:
add other extensions to the favicons list, per [caniuse](https://caniuse.com/?search=favicon) and [mdn](https://developer.mozilla.org/en-US/docs/Glossary/Favicon). uses a non-capturing group with extensions as a composite (roughly in order of likelihood, as capture groups are searched left to right).

testing against my local server, anubis appears to pass through requests to the favicon file appropriately.

draft until i update the docs and verify the regex is as tight as can be. i'll forget to file this otherwise.

issue ref: https://github.com/TecharoHQ/anubis/issues/949

Checklist:

### DIFF
--- a/cmd/anubis/main.go
+++ b/cmd/anubis/main.go
@@ -49,7 +49,7 @@ var (
 	cookieDomain             = flag.String("cookie-domain", "", "if set, the top-level domain that the Anubis cookie will be valid for")
 	cookieDynamicDomain      = flag.Bool("cookie-dynamic-domain", false, "if set, automatically set the cookie Domain value based on the request domain")
 	cookieExpiration         = flag.Duration("cookie-expiration-time", anubis.CookieDefaultExpirationTime, "The amount of time the authorization cookie is valid for")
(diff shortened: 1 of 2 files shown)


arcayr:
aside, which i'll file separately as it affects more than just this regex pattern: patterns in this file don't escape '.', which means `favicon1ico`, `sitemapAxml`, etc. all bypass anubis. needs escaping.

JasonLovesDoggo (approved):
Thanks for the catch! What's left to mark is ready?

arcayr:
changelog, updating the docs with more specifics/examples about regex matching for future users, and i want to take another look at the regex tonight to make sure it's the most optimal/sensible. i'm pretty sure it is but i wrote this over a lunch break so it was kinda done on the fly and want to be sure given it'd be the default, and the project's purpose.

i'll finalise these this evening (.au).

arcayr:
looked at the regex again, probably the sanest way to do this without getting too clever. given it's configuration and there's already documentation on the patterns, i don't think it needs anything special. rfr.

will file the other changes in a separate mr.

PR #973, opened 2025-08-14
fix(locales): remove the word "hack" from the description of Anubis

Xe:
This was causing confusion and less technical users were thinking that websites had been intruded upon, causing them to send me horrible things over email.

All non-English strings were amended using Google Translate. Please fix the localization as appropriate.

Checklist:


Earl0fPudding:
German translation is fine.
In my original translation, I even didn't use the word "Hack" and just used "Platzhalterlösung", which would translate to "placeholder solution". So, the meaning should be the same but I like your wording a bit more.

ItsHarper:
What an unfortunate disconnect in language usage. I'm sorry people sent you that stuff.

Rudxain:
Those people really need to educate themselves (in many ways).

Even the non-english Wikipedia articles about "hack" and "hacker" mention that it's not always about "breaking something".

Further resources:
- http://www.catb.org/~esr/jargon/html/H/hack.html
- https://www.gnu.org/philosophy/words-to-avoid#Hacker
- https://en.wikipedia.org/wiki/Kludge#Computer_science

PR #991, opened 2025-08-18
feat: log filtering

### LINKED ISSUE #942: Log filtering / reduction config options
Hi there, Love the project

It has helped us quite a lot since deploying! Especially with ai traffic.
We really love how expressive the CEL rules let us be. So we have elected to have anubis help us filter alot of undesirable traffic before hitting our backend. This results in an IMMENSE amount of `explicit deny` log spam, just see below. We apologize if we missed it but are there any options to disable specific logs such as explicit deny?

Image

Xe:
Closes #942 

Checklist:


data/botPolicies.yaml
```diff
@@ -132,6 +132,23 @@ dnsbl: false
...
+  # CEL log filters. Note that this is a very powerful feature and it is very easy to get
+  # yourself into trouble with this. Avoid using log filters unless you are running into
+  # circumstances like https://github.com/TecharoHQ/anubis/issues/942. This has a nonzero
+  # impact on logging, which spirals out into a more than zero impact on Anubis'
+  # performance and memory usage.
+  filters:
+    # Every filter must have a name and an expression. You can use the same expression
+    # syntax as you can with bots or thresholds.
+    - name: "http-stdlib"
+      expression: msg.startsWith("http:")
```
h-vetinari, line 149: From the peanut gallery: just from reading "filter" and "expression", I'd have no idea if messages matching the expression are filtered out or left in. From context it's likely that it's "filter out", but the underlying issue remains that "filter" is used ambiguously in many contexts - for example, a "dust filter" filters dust _out_ of air, and yet has the identical function as an "air filter" which filters air to remove pollution.

One potential solution might be an `action: keep|remove` key, even though you may only want to support removal for now. Once you allow both, you'd have to think about precedence of conflicting rules (could be the order of the filters), but the bigger issue then is that you wouldn't be able to short-circuit the _other_ expressions once a message has been filtered out, which obviously has a performance impact.

Xe: I clarified in https://github.com/TecharoHQ/anubis/pull/991/commits/0510aad9aed67ac43e81ee87340055c30c167086

PR #1005, opened 2025-08-20
fix(default-config): also block alibaba cloud

Xe:
Checklist:


Kimiblock:
This has impacted all users using Alibaba Cloud to proxy connections from China and other countries. Please consider an alternative method (such as increased workload etc.).

Xe:
This is the only option we have left. I'm sorry for the inconvenience but Alibaba cloud has not responded to the abuse contacts I've made, including through my TAM. Please contact your TAM for an explanation.

Kimiblock:
IP based blocking will probably not work. AI companies can use other ISPs or even home network. Simply blocking an IP range will not stop the desire for more data.

Kimiblock:
I'm not gonna say it's useless. But it brings more harm to end users than good. I've spent half an hour diagnosing why some users can not connect to a website I maintain, only to find they are blocked by this IP range. This is unnecessary to be the default IMO.

thess:
@Kimiblock - You find yourself in the same situation we (openwrt.org) are in given the amount of geography and ISP blocking being built into Anubis. I (personally) find this strategy a bit too punitive in nature. This, however, is not a deal-breaker. It just means more work overriding and tailoring the built-in bot-rules to suit our users and developers. For instance: There are several crawlers in addition to those from China and Russia which we want to let-in and rate-limit.

Anubis saved our main site which was, and still is, seeing massive DDoS and scraping attempts. We want to weed out the bad-guys, moderate the un-controllable and allow legit users easy access - We would absolutely not want to submit our visitors to solving word/math/puzzle games.

PR #1015, opened 2025-08-22
lib/checker: Implement X-Original-URI support

### LINKED ISSUE #453: Path regex policies ineffective when using Subrequest Authentication
I've just attempted to install anubis into an nginx-powered site using the documented Subrequest Authentication method.

What I found was that `path_regex` rules are ineffective because the subrequest does not pass any path information into anubis. When I looked through the implementation of the "check" API, I see that it does not support any way of passing path information as it just calls the reverse-proxy methods. This assumes the request URL contains the path to be checked, which it does not.

The result of this is that the path that `path_regex` rules are checking is always `/.within.website/x/cmd/anubis/api/check`

samip5:
Implemented the checking of X-Original-URI so that the path rules would work when using it with Subrequest Authentication.
Closes #453

Checklist:

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -40,6 +40,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bump AI-robots.txt to version 1.39
 - Add a default block rule for Huawei Cloud.
 - Add a default block rule for Alibaba Cloud.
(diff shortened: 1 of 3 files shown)


JasonLovesDoggo (approved):

samip5:
Do I need to do something about the spelling check?

JasonLovesDoggo:
> Do I need to do something about the spelling check?

Fixed in main! Merging this in now!

PR #1038, opened 2025-08-26
feat: add 'proof of React' challenge

Xe:
This implements a challenge method that requires the client to do a single round of SHA-256 hashing deeply nested into a Preact hook in order to prove that the client is running JavaScript. This is much more lightweight than the proof of work challenge, but also uses Preact to make sure that it runs.

I wrote this on a plane without usable documentation and only some minimal aid by [devstral 24b](https://ollama.com/library/devstral). This should be a new standard for coding interviews.

Checklist:


Xe:
Hi orange site! This is a bit of a shitpost that I wrote while on a plane. I am intending to use this to test several other meta-evasion methods for bots.

Be polite in the comments or I will remember who is ugly and prioritize support accordingly.

chris-morgan:
I’m not particularly your target audience, but I’m a bit puzzled by using something like Preact on this. There’s *nothing* about it that benefits from reactivity—rather, you’ve taken something that should be a simple bit of HTML with a little procedural JS, and made it unreasonably complicated. And it does look like you *know* you’ve made it complicated, but I can’t figure out why you would—if you were actively trying to obfuscate it (of very dubious utility anyway) I’m confident you could do much better; and I don’t think there’s much chance any kind of JS engine a bot might use could handle something simple, yet fail at your convoluted version. (If I’m wrong on that, I’d be fascinated to know what specifically tripped one up.)

As one who doesn’t use Anubis or Go, yet who got [nerd sniped](https://xkcd.com/356/) when [someone was (philosophically) wrong on the internet](https://xkcd.com/386/), I think this should be about right for the entire template (though there could be errors):

```templ
templ page(redir, challenge string, difficulty int, loc *localization.SimpleLocalizer) {
	<div class="centered-div">
		<img
			src={ anubis.BasePrefix + "/.within.website/x/cmd/anubis/static/img/pensive.webp?cacheBuster=" + anubis.Version }
			style="width:100%;max-width:256px;" />
		<p id="status">{ loc.T("loading") }</p>
		<p>{ loc.T("connection_security") }</p>
		<script>
			crypto.subtle.digest("SHA-256", new TextEncoder("utf-8").encode({{ challenge }}))
			.then(result => {
				let redir = new URL({{ redir }}, location.href);
				redir.searchParams.set("result", new Uint8Array(result).toHex());
				setTimeout(() => {
					location.href = redir.toString();
				}, {{ difficulty * 100 }});
			});
		</script>
	</div>
}
```

That seems *much* more maintainable.

I don’t know if there was a reason you were using Preact 10 instead of 11, and `@aws-crypto/sha256-js` from npm instead of the native `crypto.subtle.digest`. If it was to support ancient browsers like IE 11 (I don’t know your support target), the use of `URL.searchParams` scuppered that anyway—`crypto.subtle.digest` predates it in most browsers, with ten years of universal support. If it was to avoid the secure context requirement… again I don’t know if you care about that.

One functional change is that the pensive image, loading message and connection_security message are shown even if JS is not running. Probably good to put a `<noscript>` element in there, or similar.

If templ supports `<script type="module">` and only four years of browser support is acceptable, you could use top-level await too, `let result = await …; …` instead of `….then(result => { … })`. Honestly that would be more likely to trip up dodgy bots.

… I should go and do something else. But hey, it was fun. 🙂

Xe:
Hey @chris-morgan, thanks for the comment!

> I’m not particularly your target audience, but I’m a bit puzzled by using something like Preact on this. There’s nothing about it that benefits from reactivity—rather, you’ve taken something that should be a simple bit of HTML with a little procedural JS, and made it unreasonably complicated.

Yep, it's complicated, but it's load-bearing complexity. Using Preact here transitively checks a lot of browser compliance all at once. Doing this at least checks the following:

* DOM support
* That time is running at the normal temporal reference frame
* JSON support (some crawlers mess with JSON)
* That Preact hooks chain in the right monadic order

Half the reason I'm using a pure JS SHA256 library is to avoid the Preact hook needing to be async. I don't remember what the intersection of blue and red code is in Preact hooks, and this library is used across the tree for other things such as ensuring that Firefox has consistent performance in the proof of work rate limiter.

> I don’t know if there was a reason you were using Preact 10 instead of 11

The [preact NPM page](https://www.npmjs.com/package/preact) lists version 10 as the most recent. I can check to see if there are any major differences with version 11 though.

I'll consider this simpler approach, but I think it may be better to make things a little more complicated here as having a bigger response makes scrapers think that the page contents they got are the correct ones. This is the major advantage of Anubis: it makes the scrapers think they won.

PR #1045, opened 2025-08-26
fix(worker): constrain nonce value to be a whole integer

### LINKED ISSUE #1043: Nonce can sometimes become a non-integer value
I noticed that I was getting reliably blocked from accessing the FreeCAD forum on Firefox / macOS / arm64 for no obvious reason - the `GET pass-challenge` endpoint would fail with a `500` and the user-facing error message would only say `invalid response.`. Chromium seemed to work at first on the same machine, and on one single occasion I actually managed to get through in Firefox on private mode.

After digging a bit further, I saw that a) the issue could actually occur sporadically on Chromium as well and b) the one thing that all of the failed attempts had in common was that the `nonce` parameter was a non-integer ending in `.5` - here's the unmodified request URL that I pulled from a failed FreeCAD forum attempt:
```https://forum.freecad.org/.within.website/x/cmd/anubis/api/pass-challenge?response=00002a606101bd788836810d897858237eb007a4514e41d23106ecfdb2aef3f3&nonce=19079.5&redir=https%3A%2F%2Fforum.freecad.org%2F&elapsedTime=154```

I've then double-checked the endpoint implementation and voilà - the frontend-reported `nonce` value goes through an integer conversion, which rejects if the string is a non-integer: https://github.com/TecharoHQ/anubis/blob/main/lib/challenge/proofofwork/proofofwork.go#L44-L48

I haven't yet identified the exact location in the fronted code that generates the `nonce` value, so I can't tell yet why it ends up being a non-integer, but I'll post an update in case I'm able to find out more.

### PR DESCRIPTION
**Pull request by Xe**
Closes #1043

Sometimes the worker could get into a strange state where it has a decimal nonce, but the server assumes that the nonce can only be a whole number. This patch constrains the nonce to be a whole number on the worker end by detecting if the nonce is a decimal number and then truncating away the decimal portion.

Checklist:

### DIFF
--- a/web/js/worker/sha256-purejs.mjs
+++ b/web/js/worker/sha256-purejs.mjs
@@ -53,6 +53,18 @@ addEventListener('message', async ({ data: eventData }) => {
     nonce += threads;
     iterations++;
 
+    /* Truncate the decimal portion of the nonce. This is a bit of an evil bit
+     * hack, but it works reliably enough. The core of why this works is:
+     * 
+     * > 13.4 % 1 !== 0
+     * true
+     * > 13 % 1 !== 0
+     * false
+     */
+    if (nonce % 1 !== 0) {
+      nonce = Math.trunc(nonce);
+    }
+
     // Send a progress update from the main thread every 1024 iterations.
     if (isMainThread && (iterations & 1023) === 0) {
       postMessage(nonce);
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 <!-- This changes the project to: -->
+
 - Added a missing link to the Caddy installation environment in the installation documentation.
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: web/js/algorithms/fast.mjs [hunk 1]
```diff
@@ -4,7 +4,7 @@ export default function process(
   difficulty = 5,
   signal = null,
   progressCallback = null,
-  threads = Math.max(navigator.hardwareConcurrency / 2, 1),
+  threads = Math.trunc(Math.max(navigator.hardwareConcurrency / 2, 1)),
```
Rudxain, line 7: Why not `ceil` and remove `max`?

web/js/worker/sha256-purejs.mjs
```diff
@@ -53,6 +53,18 @@ addEventListener('message', async ({ data: eventData }) => {
...
+    /* Truncate the decimal portion of the nonce. This is a bit of an evil bit
+     * hack, but it works reliably enough. The core of why this works is:
+     * 
+     * > 13.4 % 1 !== 0
+     * true
+     * > 13 % 1 !== 0
+     * false
+     */
+    if (nonce % 1 !== 0) {
+      nonce = Math.trunc(nonce);
```
Rudxain, line 65: `x % 1 != 0` is identical to `!Number.isInteger(x)` (assuming `typeof x == "number"`), so this more complicated than an unconditional `trunc`

Issue #1067, opened 2025-08-30
bug: CI is failing on main since #1038


Xe:
- [ ] Investigate
- [ ] Fix root cause

PR #1068, opened 2025-08-31
fix(challenge/metarefresh): ensure that clients have waited long enough

Xe:
Some admins have noticed that clients are not waiting the right amount of time in order to access a resource protected by the metarefresh challenge. This patch adds a check to make sure that clients have waited at least 95% (difficulty times 950 milliseconds instead of difficulity times 1000 milliseconds) of the time they should.

If this scales, maybe time is the best way to go for Anubis in the near future instead of anything else computational.

Checklist:


Xe:
Merging on red due to #1067. I don't like this, but this is life.

PR #1191, opened 2025-10-14
Fix `ED25519_PRIVATE_KEY_HEX` character key length in docs

ghost:
Here is a proposal for a small fix where the 64 character length mentioned in the docs doesn't match the 32 one expected by anubis.

Prior to the change, from the docs:

> The hex-encoded ed25519 private key used to sign Anubis responses. If this is not set, Anubis will generate one for you. **This should be exactly 64 characters long.**

Versus what anubis expected:

```
4+02:00","level":"INFO","msg":"failed to parse and validate ED25519_PRIVATE_KEY_HEX: supplied key is not 32 bytes long, got 64 bytes"}
```

Checklist:

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- This changes the project to: -->
 
+- Fix `ED25519_PRIVATE_KEY_HEX` characters length in the documentation
 - Added `(data)/meta/default-config.yaml` for importing the entire default configuration at once.
 - Add `-custom-real-ip-header` flag to get the original request IP from a different header than `x-real-ip`.
 - Add `contentLength` variable to bot expressions.


avioletheart:
the current docs seem correct - one byte hex-encoded as ASCII is two characters:
```
~ $ openssl rand --help
Usage: rand [options] num[K|M|G|T]
[...]
 -hex                Hex encode output
[...]
 num                 Number of bytes to generate
~ $ openssl rand -hex 32
1101b52c4d5e4586b2a0e799ec6644fe1afd3629c7c4e789fdc1c905d5f00efa
~ $ openssl rand -hex 32 | wc -c
      65
# note the newline character at the end
```

ghost:
Hello,

> the current docs seem correct - one byte hex-encoded as ASCII is two characters

You are indeed right.

Sorry for the confusion, I'm gonna take a break from computers screens now. lol

Xe:
Don't worry, it happens to us all. I'll add a note about it being a 64 byte string that only contains 32 bytes of aggregate data.

PR #1217, opened 2025-10-25
fix!(policy/checker): make List and-like

Xe:
This has the potential to break user configs.

Anubis lets you stack multiple checks at once with blocks like this:

```yaml
name: allow-prometheus
action: ALLOW
user_agent_regex: ^prometheus-probe$
remote_addresses:
  - 192.168.2.0/24
```

Previously, this only returned ALLOW if _any one_ of the conditions matched. This behaviour has changed to only return ALLOW if _all_ of the conditions match.

I have marked this as a potentially breaking change because I'm absolutely certain that someone is relying on this behaviour due to spacebar heating. If this bites you, please let me know ASAP.

I used GPT-OSS 120b on local hardware to review this commit before publication as part of an experiment I will publish more about at a future date.

Checklist:

### DIFF
--- a/lib/policy/checker/checker.go
+++ b/lib/policy/checker/checker.go
@@ -20,8 +20,6 @@ type List []Impl
 // It returns true only if *all* checkers return true (AND semantics).
 // If any checker returns an error, the function returns false and the error.
 func (l List) Check(r *http.Request) (bool, error) {
-	// Assume success until a checker says otherwise.
-	allOk := true
 	for _, c := range l {
 		ok, err := c.Check(r)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


lib/policy/checker/checker.go
```diff
@@ -16,18 +16,26 @@ type Impl interface {
...
+		if !ok {
+			// One false means the combined result is false. Short-circuit
+			// so we don't waste time.
+			allOk = false
+			break
 		}
 	}
-
-	return false, nil
+	return allOk, nil
```
vikanezrimaya: Question from someone not proficient in Go: Why not just `return false, nil` instead of using a variable and then returning it, and returning at the bottom too?

Xe: Both are valid! I think it may be better to be overly explicit though, see https://github.com/TecharoHQ/anubis/pull/1217/commits/5a8df9d6885b61a2ecafc855a6e2f3399736e0fb

Issue #1235, opened 2025-11-02
Anubis writes gigabytes to /var/log/messages


Midar:
It seems every single request is logged to /var/log/messages in JSON form for some reason. I've essentially traded a DDoS on my CPU for a DDoS on my SSD. Config is the default.

Xe:
Which distribution are you using?

Midar:
Fedora 42 with Anubis installed from repos.

Midar:
I just noticed that all those have "level":"INFO", so I'm guessing the default level for Anubis when none has been specified is INFO. Maybe this should be something more conservative?

Xe:
I'm not seeing the file /var/log/messages on my Fedora workstation. Either way, please file a bug with the Fedora packaging. I'm not sure what they changed from upstream.

Midar:
Fedora Server in this case (now just updated to 43). It seems that it's set up so that the journal gets written in plain text to /var/log/messages. You should see the same using journalctl on Workstation.

Xe:
Try setting the environment variable `SLOG_LEVEL` to `warn` in the service's unit configuration.

Midar:
The environment seems to all come from the .env file. I'll add it there, thx. But still, shouldn't warn be the default if info is so noisy?

Xe:
Unfortunately, people's fail2ban pipelines rely on the noise. I am understandably reluctant to break them. I also personally tend to have services be very noisy in the logs because storage is cheap and I'd much rather see instant evidence of a service working than having to guess.

Midar:
Setting it to warn brought it down to reasonable levels.

Can you maybe add a warning in the documentation as this very quickly accumulated multiple gigabytes?

confor:
i'm seeing similar behavior in debian, with `/var/log/journal` going almost 1 gb in about 15 days.

i understand why the default loglevel is so verbose, but it would be nice to have a warning as Midar says, along with steps to configure it down to `warn` or `err`. now i've been pruning the logs regularly but the first time i noticed, my /var folder was 4.5 gb.

for anyone running systemd, clear some space with:
```
$ sudo journalctl --vacuum-time=10d
```

Xe:
I'm adding logging options in https://github.com/TecharoHQ/anubis/pull/1299. Would this be sufficient for your needs?

Bracket-I-I:
You cannot expect anyone queer to solve the underlying problem. The root problem requires antisemitism against ashkenazim. 
[...]

Issue #1009, opened 2025-08-21
"client was given a challenge but does not in fact support gzip compression"


W1zzardTPU:
Should this be an error or some lower loglevel?

It's cluttering up my logs from all the terribly programmed bots ;)

PR #1299, opened 2025-11-20
feat: writing logs to the filesystem with rotation support

### LINKED ISSUE #1125: feature: Do not log IP address
**Is your feature request related to a problem? Please describe.**
For privacy reasons, we do not want to permanently store IP addresses in hour logs.

**Describe the solution you'd like**
An option to have X-Real-IP redacted from the syslog output.

**Describe alternatives you've considered**
Not setting X-Real-IP at all, but that gives anubis less information to act on

**Additional context**
Syslog lines include ``"x-real-ip":"2804:d55:…"``

Xe:
Closes #1125

When I designed Anubis' logging subsystem details, I did it with the experience of a recovering former SRE under the axiom of "you pay for log storage (cheap) or you pay for extended outages (expensive)". Most of the environments I run Anubis in have a rotating 10 megabyte log window with minimal retention.

Apparently this causes issues when used in normal distributions with normal distribution packaging. journald has issues handling Anubis' fairly moderate amount of logging and there's been [numerous](https://github.com/TecharoHQ/anubis/issues/1235) [comments](https://github.com/TecharoHQ/anubis/issues/1009) about this.

I concede that maybe an alternate approach is required here. The horrible part is that changing the default logging level isn't really an option because that would break automation.

This PR adds the ability to write logs to the filesystem instead of standard error. In order to configure this, add a `logging` block to your policy file like this:

```yaml
logging:
  sink: file
  parameters:
    file: "./var/anubis.log"
    maxBackups: 3 # keep at least 3 old copies
    maxBytes: 67108864 # each file can have up to 64 Mi of logs
    maxAge: 7 # rotate files out every n days
    compress: true
    useLocalTime: false # timezone for rotated files is UTC
```

This makes Anubis log everything to `./var/anubis.log` until the log file is either 7 days old or 64Mi big. When a file gets too big, it gets compressed with gzip and renamed to the date it was rotated out.

Additionally, you can also set the log level in the policy file:

```yaml
logging:
  sink: stdio
  level: warn
```

The format of `level` is what is allowed by [`slog.Level`](https://pkg.go.dev/log/slog#Level) in the Go standard library. Typical values include:

| Log level | Use in Anubis |
| :-- | :-- |
| `debug` | The raw unfiltered torrent of doom. Only use this if you are actively working on Anubis or have very good reasons to use it. |
| `info` | The default logging level, fairly verbose in order to make it easier for automation to parse. |
| `warn` | A "more silent" logging level. Much less verbose. Some things that are now at the `info` level need to be moved up to the `warn` level in future patches. |
| `error` | Only log error messages. |

These are also the same levels accepted by `--slog-level` and `SLOG_LEVEL`.

Once this is merged, #991 needs to be fixed up and amended.

Checklist:

- [x] Add documentation for this feature
- [x] Add documentation about Anubis' logging verbosity and how to change it
- [x] Add smoke test that actually makes sure logs are written to the log file

### DIFF
--- a/docs/docs/admin/policies.mdx
+++ b/docs/docs/admin/policies.mdx
@@ -372,15 +372,14 @@ This isn't currently used by Anubis, but will be in the future for "slightly imp
 
 The `file` sink makes Anubis write its logs to the filesystem and rotate them out when the log file meets certain thresholds. This logging sink takes the following parameters:
 
-| Name                | Type                  | Example               | Description                                                                                                                                                                                                                                 |
-| :------------------ | :-------------------- | :-------------------- | :------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
-| `file`              | string                | `/var/log/anubis.log` | The file where Anubis logs should be written to. Make sure the user Anubis is running as has write and file creation permissions to this directory.                                                                                         |
-| `maxBackups`        | number                | `3`                   | The number of old log files that should be maintained when log files are rotated out.                                                                                                                                                       |
-| `maxBytes`          | number of bytes       | `67108864` (64Mi)     | The maximum size of each log file before it is rotated out.                                                                                                                                                                                 |
-| `maxAge`            | number of days        | `7`                   | If a log file is more than this many days old, rotate it out.                                                                                                                                                                               |
-| `oldFileTimeFormat` | Go time format string | `2006-01-02T15-04-05` | A [Go time format string](https://pkg.go.dev/time#Layout) that describes how the filenames of old log files should be constructed. This default somewhat conforms to [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339).                    |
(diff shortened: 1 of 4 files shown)


cmd/anubis/main.go
```diff
@@ -469,15 +476,15 @@ func main() {
 
 	var h http.Handler
 	h = s
+	h = internal.JA4H(h)
 	h = internal.CustomRealIPHeader(*customRealIPHeader, h)
 	h = internal.RemoteXRealIP(*useRemoteAddress, *bindNetwork, h)
 	h = internal.XForwardedForToXRealIP(h)
 	h = internal.XForwardedForUpdate(*xffStripPrivate, h)
-	h = internal.JA4H(h)
```
JasonLovesDoggo, line 476: Any reason for this swap?

Xe: That was something I forgot to make its own PR.

PR #132, opened 2025-03-26
Add middleware to set Cache-Control header for challenge HTML

JasonLovesDoggo:
Added a Cache Control header to the challenge page to hopefully stop tools from caching it's content (e.g. a js file)

see https://github.com/TecharoHQ/anubis/issues/31#issuecomment-2744414745 & https://github.com/TecharoHQ/anubis/issues/71#issuecomment-2745445389

diff (taken from `curl -i -A "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36" localhost:8923`)

```diff
 HTTP/1.1 200 OK
+Cache-Control: no-store
 Content-Type: text/html; charset=utf-8
 Set-Cookie: within.website-x-cmd-anubis-auth=; 
 Expires=Wed, 26 Mar 2025 21:39:43 GMT; Max-Age=0; SameSite=Lax
 Date: Wed, 26 Mar 2025 22:00:00 GMT
 Transfer-Encoding: chunked
```

Checklist:

- [x] Added a description of the changes to the `[Unreleased]` section of docs/docs/CHANGELOG.md
- [x] Tested this at least manually

PR #1474, opened 2026-02-22
feat(http): set "Cache-Control: no-store" on error responses

gucci-on-fleek:
Since #132, Anubis has set `Cache-Control: no-store` on challenge responses. However, this does not apply to deny responses, meaning that if Anubis is configured to block certain user agents and is behind a caching reverse proxy, this error page will be cached and served to all subsequent requests, even those with an allowed user agent. This commit configures the error page responder to also set the `Cache-Control` header, meaning that deny and challenge responses will now both have the same behaviour.

Disclosure: I let GitHub Copilot write most of the testing code for me. I didn't use an agent or anything similar, I just pressed <kbd>Tab</kbd> a few times in the test section, and then carefully reviewed the code myself. I didn't accept any completions in any of the other files. This isn't an agent, so I don't think that I need to add a commit trailer for this, but please let me know if you want me to edit my commits and add that in.

Checklist:

### DIFF
--- a/.github/workflows/docker-pr.yml
+++ b/.github/workflows/docker-pr.yml
@@ -31,7 +31,7 @@ jobs:
           node-version: "24.11.0"
       - uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
         with:
-          go-version: "1.25.7"
+          go-version: "stable"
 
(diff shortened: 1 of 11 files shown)


Xe (approved):

gucci-on-fleek:
I've pushed 2 new chore commits, and the CI passes now.

PR #1622, opened 2026-05-11
feat: add Caddy middleware module

### LINKED ISSUE #16: A module for Caddy
It would be nice if this could be possible to be built directly as a Caddy module (since both are Go programs), this way it could be easily used within it (as request middleware), without complex/error-prone/inefficient request routing to an external process. (and thus more adoption, since it'd be very straightforward)

Ideally it'd allow reconfiguring options, both defaults on global level and per-path, this way people would have full freedom of choice on which paths to enable Anubis if not on all, and fine-tune some settings on some specific ones if needed.

As an inspiration for similar approach taken by another project, look at: https://github.com/darkweak/souin/tree/master/plugins/caddy

Souin cache is structured as a library, but offers plugins, including Caddy, to directly use it within the web servers (though in its case, the docs are a bit lacking), with the file structure that it has, it can easily be added to Caddy with

```
xcaddy build --with github.com/darkweak/souin/plugins/caddy
```

Other examples include: `github.com/dunglas/mercure/caddy`, `github.com/dunglas/vulcain/caddy`

yuganoffl:
## Summary
- add a Caddy HTTP middleware module under `plugins/caddy`
- add `Server.HandlerWithNext` so middleware integrations can safely inject downstream handlers without mutating shared server state
- document basic `xcaddy` build and Caddyfile usage

Closes #16.

## Validation
- `go test ./lib -run TestHandlerWithNextDoesNotMutateServer`
- `(cd plugins/caddy && go test ./...)`
- `git diff --check`

Note: the full pre-commit asset build fails on this macOS environment because `web/build.sh` uses `shopt -s globstar`, which is unavailable in the system bash here. The generated `web/static/app.js` step completed before that failure, and the targeted Go tests above pass.


yuganoffl:
Small update: I registered the `anubis` Caddyfile handler directive and added parser tests, so the documented Caddyfile example now round-trips through the plugin.

I’m intentionally keeping this PR scoped to the minimal native middleware path: one small public lib helper, a nested Caddy module, focused Caddyfile parsing, and docs. Validation:

- `go test ./lib -run TestHandlerWithNextDoesNotMutateServer`
- `(cd plugins/caddy && go test ./...)`
- `git diff --check`

Xe:
This is an AI slop PR that won't work because of `go generate`d files. I'm going to contact the bounty service and ask that Anubis be removed from that listing.

PR #1637, opened 2026-05-19
perf: avoid fmt.Sprintf and int round-trip in PoW validation

jvoisin:
Use direct string concatenation (`challenge + nonceStr`) instead of `fmt.Sprintf("%s%d", challenge, nonce)`, and keep nonce as a string since strconv.Atoi was only used to feed the integer back into Sprintf. The Atoi call is retained for input validation.

Benchmark results (arm64, -benchmem -count=5):

  Old (fmt.Sprintf):  ~284-328 ns/op, 160-168 B/op, 2-3 allocs/op
  New (string concat): ~92-94 ns/op,  144 B/op,     1 alloc/op

Checklist:


jvoisin:
Piled on top of #1639

PR #1638, opened 2026-05-19
perf(policy): cache Bot.Hash() to remove per-request rehashing

jvoisin:
Bot.Hash() was recomputing
the hash of every checker on every call, even though the inputs (Bot.Name and Bot.Rules) are immutable once ParseConfig ran.

This change adds an unexported `hash` field to Bot, populates it once at parse time inside ParseConfig, and has Hash() return it directly when set. When the cache is empty the function falls back to recomputing the digest on demand, so the current callers that build Bots inline do not have to know about it.

This implementation was picked instead of a lazy initialization on first call, a sync.once, or NewBot() or Seal() constructs as it's the least invasive one.

Performances-wise, this change removes a bunch of allocations per requests thanks to the cache, and makes the computation cheaper as well by removing an fmt.Sprintf call.

Checklist:


jvoisin:
Piled on top of #1639

PR #1639, opened 2026-05-19
perf(lib): iterate s.policy.Bots by index to drop per-call heap copy

jvoisin:
`s.check()` walks every bot rule on every request. The previous `for _, b := range s.policy.Bots` loop copied every Bot into a loop-local b and, because the match branch returns &b, the compiler moves b to the heap, generating one Bot-sized heap allocation per check() call plus a struct copy on every iteration.

This commit switches to an index based iteration and takes the address of the slice element directly, which is alright as s.policy.Bots is populated once in policy.ParseConfig and never changed afterwards.

Performances-wise, this removes a per-iteration heap allocation and Bot copy, which isn't negligible when the number of rules starts to grow.

Checklist:

- [ ] Added test cases to [the relevant parts of the codebase](https://github.com/TecharoHQ/anubis/blob/main/CONTRIBUTING.md)

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -30,6 +30,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Add config option to add ASN to logs/metrics.
 - Log weight when issuing challenge
 - Fix `path_regex` and CEL `path` rules not matching when using Traefik `forwardAuth` middleware. Anubis now checks `X-Forwarded-Uri` (Traefik) in addition to `X-Original-URI` (nginx) when resolving the request path in subrequest mode ([#1628](https://github.com/TecharoHQ/anubis/issues/1628)).
+- Marginally increase the performances of requests processing
+- Marginally improve the performances of PoW validation
 
 ## v1.25.0: Necron
 
--- a/internal/hash.go
+++ b/internal/hash.go
@@ -11,9 +11,8 @@ import (
 // SHA256sum computes a cryptographic hash. Still used for proof-of-work challenges
 // where we need the security properties of a cryptographic hash function.
 func SHA256sum(text string) string {
-	hash := sha256.New()
-	hash.Write([]byte(text))
-	return hex.EncodeToString(hash.Sum(nil))
+	sum := sha256.Sum256([]byte(text))
+	return hex.EncodeToString(sum[:])
 }
 
 // FastHash is a high-performance non-cryptographic hash function suitable for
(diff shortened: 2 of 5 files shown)


lib/anubis.go
```diff
@@ -682,7 +689,7 @@ func (s *Server) check(r *http.Request, lg *slog.Logger) (policy.CheckResult, *p
 		if match {
 			switch b.Action {
 			case config.RuleDeny, config.RuleAllow, config.RuleBenchmark, config.RuleChallenge:
-				return cr("bot/"+b.Name, b.Action, weight), &b, nil
+				return cr("bot/"+b.Name, b.Action, weight), b, nil
```
JasonLovesDoggo: Can we return a cloned bot on match instead of `&s.policy.Bots[i]`? `hydrateChallengeRule()` mutates `*policy.Bot`, so this now writes into shared policy state.

jvoisin: Huh, good point. I don't think it's worth the complexity of adding a deep-copy method for this. Shall I simply drop the commit?

JasonLovesDoggo: Probably the easiest option. 

Also a small nit that I'm not going to block on is that, given the rule hash caching, we now have the possibility of stale hashes. However, I don't believe we update Rule.Action/other params anywhere despite them being Public values.

JasonLovesDoggo:
hey @jvoisin this all looks great! Would you mind updating to the commits to [be signed](https://anubis.techaro.lol/docs/developer/signed-commits/). 

Additionally, did you have any benchmarks of these changes? All good if not, just curious.

Besides that LGTM!

JasonLovesDoggo (approved):

jvoisin:
The SHA256 one:

Bench	sec/op	B/op	allocs/op
SHA256_PolicyInputs	−9.7%	−18.8%	−33% (3→2)
SHA256_ChallengeInputs	−8.7%	−10.9%	−25% (4→3)
SHA256_BotRuleInputs	−8.8%	−17.0%	−33% (3→2)
SHA256_CELInputs	−7.3%	−14.0%	−25% (4→3)
SHA256_ASNInputs	−13.0%	−15.4%	−25% (4→3)
SHA256_PolicyList	−3.8%	−8.4%	−21% (42→33)
geomean	−8.6%	−14.2%	−27.3%

The PoW validation one:

	sec/op	B/op	allocs/op
Validate	−24.1% (923ns → 700ns)	−29.2% (192→136)	−50% (6→3)

And the Bot hash one (note that the cache isn't exercised in the benchmark, only the removal of fmt.Sprintf):

	sec/op	B/op	allocs/op
BotHash	−19.2% (1135ns → 917ns)	−18.6% (344→280)	−25% (12→9)

PR #1670, opened 2026-06-06
fix: allow local IP addresses in X-Forwarded-For

### LINKED ISSUE #1668: XForwardedForToXRealIP cannot handle link-local addresses
### Describe the bug

Another issue related but distinct from #1667 

If I *don't* pass the X-Real-IP headers that Anubis doesnt like, and it tries to copy from the xff header, it fails when the address is not publicly routable

This seems to be intended behavior of the xff module for whatever reason. It just returns an empty string for local addresses. https://github.com/sebest/xff/blob/671bd2870b3a94230620f8a68525a3a17e3a4585/xff.go#L52

### Steps to reproduce

1. Access Anubis through a reverse proxy that sets `X-Forwarded-For` but not `X-Real-IP` from a link-local address
2. Anubis responds with a 500 error

### Expected behavior

Anubis forwards traffic as expected

### Your operating system and its version.

Fedora 42

### Your browser and its version.

Any

### Additional context

Log messages with debug logging: 
```
Jun 05 05:08:43 offworld.city anubis[1591789]: {"time":"2026-06-05T05:08:43.498053638Z","level":"DEBUG","source":{"function":"main.main.XForwardedForToXRealIP.func5","file":"/home/drewcassidy/anubis/internal/headers.go","line":101},"msg":"setting X-Real-Ip from X-Forwarded-For","to":"","x-forwarded-for":"fe80::9000:6ff:fe59:27ac"}
Jun 05 05:08:43 offworld.city anubis[1591789]: {"time":"2026-06-05T05:08:43.498498541Z","level":"ERROR","source":{"function":"github.com/TecharoHQ/anubis/lib.(*Server).maybeReverseProxy","file":"/home/drewcassidy/anubis/lib/anubis.go","line":205},"msg":"check failed","subsystem":"anubis","host":"git.offworld.city","method":"GET","path":"/api/v1/user","user_agent":"Go-http-client/2.0","accept_language":"","priority":"","x-forwarded-for":"fe80::9000:6ff:fe59:27ac","x-real-ip":"","err":"[misconfiguration] X-Real-Ip header is not set"}
```

drewcassidy:
The xff library returns an empty string when it encounters a private address in X-Forwarded-For. This means that when X-Real-IP is unset, it gets set to an empty string, tripping a confusing error message from Anubis. I figure if the data in the XFF header is invalid in some way, its better to see that explicitly as a failure to parse X-Real-IP rather than wondering why its an empty string

The only logic the xff.Parse function did was check the IP and strip whitespace. So I replaced it with just stripping whitespace. 

Fixes #1668 

Checklist:

- [ ] Added test cases to [the relevant parts of the codebase](https://github.com/TecharoHQ/anubis/blob/main/CONTRIBUTING.md) (I have no idea what this would be)

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Marginally improve the performances of challenges generation/display
 - Significantly improve the performances of the gzip middleware
 - Significantly improve the performances of the PoW validation
+- Fix local addresses being rejected in the X-Forwarded-For header when X-Real-IP is unset ([#1668](https://github.com/TecharoHQ/anubis/issues/1668))
 
 ## v1.25.0: Necron
 
--- a/go.mod
+++ b/go.mod
@@ -24,7 +24,6 @@ require (
 	github.com/playwright-community/playwright-go v0.5200.1
 	github.com/prometheus/client_golang v1.23.2
 	github.com/redis/go-redis/v9 v9.17.2
-	github.com/sebest/xff v0.0.0-20210106013422-671bd2870b3a
 	github.com/shirou/gopsutil/v4 v4.25.11
 	github.com/testcontainers/testcontainers-go v0.40.0
 	go.etcd.io/bbolt v1.4.3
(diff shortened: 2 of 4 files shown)


Xe:
This should probably be gated by a configuration setting as a common attack vector for webapps is to do X-Forwarded-For / X-Real-Ip smuggling.

PR #1694, opened 2026-06-17
perf(internal/gzip): share writer pool across middleware reconstructions

jvoisin:
GzipMiddleware pooled its gzip.Writer in a closure-local sync.Pool (#1654). That works only when the middleware is built once, but its sole call site (RenderIndex) reconstructs it on every request because the wrapped handler embeds the request-specific challenge page. The pool was therefore created empty, used once, and garbage collected on every challenge render, never reusing a writer. The setup-time gzip.NewWriterLevel validation allocated a second throwaway ~1.18 MiB writer per request on top of that.

This commit hoists the pools to a process-global sync.Map keyed by compression level so the deflate buffers survive reconstruction, and validate the level with a cheap range check (gzip.HuffmanOnly..BestCompression) instead of allocating a writer to probe for an error.

Measured with the middleware constructed per request, which is the production pattern:

  B/op       1,210,345 -> 1,930   -99.84%
  allocs/op  37        -> 14      -62%
  ns/op      ~31µs     -> ~5.9µs  ~-81%

To prove that this fixes a real issue, tests are added, as the middleware was untested:
- gzip round-trip and non-gzip passthrough correctness
- TestGzipMiddlewareInvalidLevel: construction-time range-check panic
- TestGzipMiddlewarePoolSurvivesReconstruction: asserts via testing.AllocsPerRun that per-request construction stays ~14 allocs, not the ~37 of a defeated pool
- BenchmarkGzipMiddleware: per-request construction pattern

Checklist:

### DIFF
--- a/internal/gzip.go
+++ b/internal/gzip.go
@@ -2,27 +2,47 @@ package internal
 
 import (
 	"compress/gzip"
+	"fmt"
 	"io"
 	"net/http"
 	"strings"
 	"sync"
 )
 
-func GzipMiddleware(level int, next http.Handler) http.Handler {
-	// Validate the level once at setup; gzip.NewWriterLevel only fails for
-	// invalid levels and we'd rather panic now than mid-request.
-	if _, err := gzip.NewWriterLevel(io.Discard, level); err != nil {
-		panic(err)
-	}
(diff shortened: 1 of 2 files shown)


jvoisin:
Full disclosure: this is fixing a regression I introduced :P

Xe (approved):

PR #1697, opened 2026-06-20
Use system font stack instead of custom fonts

### LINKED ISSUE #142: Use system font instead of custom one
Don't really need to load a font in the css, the system `sans-serif` font should be fine

noahkhomer18:
Closes #142

### DIFF
--- a/xess/static/podkova.css
+++ b/xess/static/podkova.css
@@ -1,7 +1,0 @@
-@font-face {
-  font-family: "Podkova";
-  font-style: normal;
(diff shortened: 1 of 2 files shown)


Xe:
The custom fonts are there for a reason: it makes the page bigger which makes some scrapers think the Anubis challenge page is the content they want.

noahkhomer18:
Ah, that makes sense — the oversized fonts trick scrapers into thinking the challenge page is real content. Good to know! Closing this PR.

Issue #1731, opened 2026-07-07
Cookies disabled page provides no way to get to destination


svensven:
### Describe the bug

The "Oh noes!" page only has a "Go home" link which goes to the site root, and refreshing after enabling cookies doesn't navigate to the desired page either, and there's no back history, so the only option is to close the tab and find the original link to open the tab again.

### Steps to reproduce

1. disable first-party cookies
2. go to https://docs.fedoraproject.org/en-US/fedora/latest/release-notes/desktop/
3. get "oh noes" page
4. enable cookies for the site (as the page suggests - "Please enable cookies for this domain.")
5. fail to have any way to actually get to the intended page

### Expected behavior

Link to the intended page, or make refresh work, or be able to click back to get there.

### Your operating system and its version.

Windows 10

### Your browser and its version.

Firefox 152

marcusxss:
Was just about to request this as well. It's suddenly very hard to get to the destination because of this redirect, and I expect that's just something that generally hits human visitors and not so much bots? I could be wrong, though.

Edit: Possibly related #1658

Chang-Jin-Lee:
Root cause: in `PassChallenge` (`lib/anubis.go`), the missing-test-cookie check runs *before* the redirect is parsed into `r.URL`/validated against `RedirectDomains`, and the cookies-disabled error page (`web/index.templ`'s `errorPage`) never receives the destination at all — it only ever links to `/`. So once you land on that page there's no way back to where you were going, even after enabling cookies.

Proposed fix (draft, happy to adjust based on feedback):

- Reorder `PassChallenge` so the cookie check runs *after* the existing scheme + `RedirectDomains`/host validation, instead of before. This means the value handed to the error page has already passed the same open-redirect checks as the success path — no new validation logic, just moved.
- Add an optional `retryURL` param to `errorPage`/`web.ErrorPage`, rendered as a "Try again" link (reusing the existing `try_again` locale string, which is already translated in all locales but currently unused anywhere).
- New `respondCookiesDisabled` helper in `lib/http.go` builds that page with the validated destination; all other call sites of `respondWithError`/`respondWithStatus` are untouched (pass `""` for the new param), so this doesn't touch the other ~29 call sites.

[...]

PR #1752, opened 2026-07-16
fix(web): preserve redirect destination on cookies-disabled error page

Chang-Jin-Lee:
## Summary

- `PassChallenge` checked for the missing Anubis test cookie *before* the redirect target was validated against the existing scheme/`RedirectDomains` checks, and the cookies-disabled error page never received a destination at all — it only ever linked to `/`. Once a visitor with cookies disabled landed on that page there was no way back to where they were going, even after enabling cookies (#1731).
- Moved the cookie check to run *after* the existing scheme + `RedirectDomains`/host validation, so the destination handed to the error page has already passed the same open-redirect checks as the success path (no new validation logic added, just reordered).
- Added an optional `retryURL` param to `errorPage`/`web.ErrorPage`, rendered as a "Try again" link back to the destination, reusing the existing `try_again` locale string (already translated in every locale file, previously unused).
- Added a `respondCookiesDisabled` helper in `lib/http.go` for this one call site; the other ~29 `respondWithError`/`respondWithStatus` call sites are untouched (pass `""` for the new template param).

Side effect worth flagging for review: previously, disabled cookies + a disallowed redirect domain would show the cookies-disabled message; after the reorder it shows "redirect domain not allowed" instead, since that check now runs first. This seemed like the right call security-wise (never construct a retry link before the destination is vetted) — flagged in the issue thread before opening this PR, see https://github.com/TecharoHQ/anubis/issues/1731#issuecomment-4991462127.

## Test plan

- [x] `go build ./...`
- [x] `SKIP_INTEGRATION=1 go test ./...` (full suite, including the pre-existing `TestPassChallengeXSS` open-redirect/XSS regression tests)
- [x] Added `TestPassChallengeCookiesDisabledPreservesDestination` and `TestPassChallengeCookiesDisabledStillEnforcesRedirectDomain` in `lib/anubis_test.go`
- [x] `go vet ./...`, `go tool staticcheck ./...`, `go tool govulncheck ./...`
- [x] `cspell` on changed files
- [ ] `npm run test:integration` (Playwright) — could not be exercised in my sandboxed environment due to headless-browser dependency/sandbox constraints; confirmed via a control run that this failure is pre-existing and occurs identically on unmodified `main`, and that suite covers bot-fingerprinting detection rather than the cookie/redirect path touched here. Flagging so CI can confirm.

Assisted-by: Claude Sonnet 5 via Claude Code

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 <!-- This changes the project to: -->
 
+- Fix the cookies-disabled error page having no way to reach the original destination by adding a "Try again" link back to the (already validated) redirect target ([#1731](https://github.com/TecharoHQ/anubis/issues/1731))
+
 ## v1.26.0-pre2
 
 - Add `(data)/clients/google-user-triggered-fetchers.yaml` snippet that allows Google-owned user-triggered fetchers (Google Translate's website translation proxy, Google Read Aloud, Google Messages link previews) by their [published IP ranges](https://developers.google.com/static/crawling/ipranges/user-triggered-fetchers-google.json), fixing the infinite challenge loop for visitors using Google Translate ([#444](https://github.com/TecharoHQ/anubis/issues/444))
--- a/lib/anubis.go
+++ b/lib/anubis.go
@@ -523,14 +523,6 @@ func (s *Server) PassChallenge(w http.ResponseWriter, r *http.Request) {
 		cookiePath = strings.TrimSuffix(anubis.BasePrefix, "/") + "/"
 	}
 
-	if _, err := r.Cookie(anubis.TestCookieName); errors.Is(err, http.ErrNoCookie) {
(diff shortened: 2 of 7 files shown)


Chang-Jin-Lee:
Closing this — half of it landed upstream and the other half is better redone than rebased.

The reordering this PR was built around is already on `main`. `PassChallenge` now calls `s.validateRedirect(redir)` first and rejects there, so by the time the test-cookie check runs at `lib/anubis.go:603` the destination has passed scheme and `RedirectDomains` validation. That is what "Share redirect validation between challenge completion and subrequest authentication" did, and it makes the inline validation block in this branch redundant.

What is still missing is narrower: line 607 hands the error page an empty destination, so the cookies-disabled page continues to link only to `/`, which is what #1731 reports. I have left the details there.

Rebasing this branch is not the way to get that. It is 97 commits behind, the conflict is in `web/index_templ.go`, and `templ` is generated code I cannot regenerate or compile on this machine — hand-resolving 19 hunks of it and force-pushing would be a guess. A fresh, much smaller change on top of the refactor is the right shape.

PR #1797, opened 2026-08-04
chore(cmd/anubis): enable Partitioned flag by default

Xe:
This is probably gonna bite someone somewhere, but I am told that this fixes the random Anubis cookie rejections. I am not able to reproduce this reliably despite many months of trying and assurances that this default configuration is broken.

Let's see how it fails in the real world, I guess.

Closes: #1701

Checklist:

- [x] Added a description of the changes to the `[Unreleased]` section of docs/docs/CHANGELOG.md
- [ ] Added test cases to [the relevant parts of the codebase](https://github.com/TecharoHQ/anubis/blob/main/CONTRIBUTING.md)
- [x] Ran integration tests `npm run test:integration` (unsupported on Windows, please use WSL)
- [x] All of my commits have [verified signatures](https://anubis.techaro.lol/docs/developer/signed-commits)

Issue #1701, opened 2026-06-24
User blocked by default when using an in-app link.


uplg:
### Describe the bug

To access a website protected by Anubis using latest Vanadium from GrapheneOS. I get a "Oh noes", and I'm enforced to "enable Third party cookies".

This is. wrong.

### Steps to reproduce

1. Access https://lore.kernel.org/lkml/CAHk-=wghMm2c+AYEcwYY7drSVXB27DYqc-ZXpFiq=XFs-w59wA@mail.gmail.com/ using Vanadium latest (149.0.7827.159)
2. See that you are blocked and enforced to enable "Third party cookies" to get access.

### Expected behavior

- Should allow access and don't enforce "third party cookies".

### Your operating system and its version.

Android 17

### Your browser and its version.

Vanadium 149.0.7827.159

Xe:
This is expected behaviour.

uplg:
> This is expected behaviour.

May I know the reasoning behind that?

Xe:
Vanadium is miscategorizing Anubis' **first-party** cookies as third party cookies. Please file this as a bug with Vanadium.

avioletheart:
fwiw i can't reproduce this on the same version of vanadium on grapheneos with default settings (third-party cookies blocked), i can get through the anubis check.

uplg:
> fwiw i can't reproduce this on the same version of vanadium on grapheneos with default settings (third-party cookies blocked), i can get through the anubis check.

Right, I also can't reproduce now, but reproduced more than 3 times before. And some others on reddit had the same issue. Not sure what happens here.

Xe:
This happens every so often with Vanadium. Please file a bug upstream.

uplg:
I've filed GrapheneOS/Vanadium#1171. That said, I don't think "Vanadium is miscategorizing a first-party cookie" is the full picture, and I'd ask you to reconsider closing this as purely a downstream bug.

Anubis isn't passively setting a neutral first-party cookie that Vanadium then mislabels. It **explicitly tags the cookie `SameSite=None`** by default. Verified against this very deployment (lore.kernel.org, Anubis 1.25.0):

```
set-cookie: techaro.lol-anubis-auth=; Path=/; Max-Age=0; Secure; SameSite=None
set-cookie: techaro.lol-anubis-cookie-verification=<uuid>; Path=/; Secure; SameSite=None
```

And that's the shipped default:

```go
// cmd/anubis/main.go
[...]

PR #1798, opened 2026-08-04
fix(lib): rename cookies based on settings

Xe:
Cookies are horrible: a rant by Xe Iaso.

Cookies are horrible. When cookies are sent from server to client, they include a number of settings that I mistakenly let administrators control. Turns out that when the browser stores them, they uniquely identify cookies by the combination of both name and settings. These settings are not sent back to the server upon requests. As a result, if any administator changes any cookie setting without changing the cookie name, the client sends two cookies. This is not good because the server tries to clear the errant cookie, but the client discards that because it doesn't match what the server sent.

To fix this, cookies are now dynamically renamed based on cookie settings. This behaviour cannot be disabled.

Ref: #1797
Ref: #1701

Checklist:

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -11,11 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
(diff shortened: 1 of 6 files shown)


ThiefMaster:
Including the hash in the cookie name breaks the advanced haproxy setup since it's now no longer (easily) possible to know what cookie to look for.

PR #1821, opened 2026-08-06
fix(localization): don't load manifest.json as a locale

### LINKED ISSUE #1776: "making_sure_not_bot" not found in language "und"
### Describe the bug

this only seems to affect certain bots so not a high priority issue but we are seeing a low volume of unhandled exceptions with a stack trace in our logs that results in a server disconnected error on the other side. the error in the anubis logs is:
```
http: panic serving xxx.xxx.xxx.xxx:xxxxx: message "making_sure_not_bot" not found in language "und"
```
sure enough, the app logs show server disconnected errors associated with subrequests to our anubis deployment that include
```
Accept-Language: und;q=0.9
```

### Steps to reproduce

1. set `Accept-Language` to some nonsense value like `und`
2. get challenged by Anubis

### Expected behavior

localize to a default locale when Accept-Language is malformed or not present

### Your operating system and its version.

N/A

### Your browser and its version.

N/A

### Additional context

_No response_

SihyeonJeon:
`Accept-Language: und` panics the challenge handler.

`NewLocalizationService` loads every `locales/*.json` into the bundle, including `manifest.json`. go-i18n derives a locale from the file name, so `manifest` — which is not a language tag — registers under `language.Und`. A request whose highest priority tag is `und` then matches that entry instead of the `"en"` fallback appended in `GetLocalizerFromRequest`, the message is missing from it, and `MustLocalize` panics:

```
http: panic serving 10.0.0.1:1234: message "making_sure_not_bot" not found in language "und"
```

Skipping `manifest.json` when populating the bundle removes the bogus locale. Other unusual tags already fell back correctly — `und-US`, `mul`, `zxx`, `qaa`, and unknown subtags such as `xx` — only the exact zero tag was affected.

The test asserts through `T()` rather than `GetLang()` on purpose: `GetLang()` swallows the lookup error and reports `"en"` either way, so a test built on it would pass with and without this change. Reverting only `localization.go` makes it fail with the panic above.

Fixes #1776

Checklist:

`npm run test`, `go vet ./...` and `staticcheck` all pass. I could not get the Playwright harness up locally: `TestPlaywrightBrowser` fails with `could not connect to remote browser: Connection closed` on unmodified `main` as well, so it looks environmental rather than related to this change.

### DIFF
--- a/docs/docs/CHANGELOG.md
+++ b/docs/docs/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Updates IP list for DuckDuckBot ([#1810](https://github.com/TecharoHQ/anubis/pull/1810))
 - Update Huawei Cloud IP list ([#1814](https://github.com/TecharoHQ/anubis/pull/1814))
 - Add standard library rules for making Dillo less threatening to Anubis and [document how to enable them](./admin/faq.mdx#how-do-i-allow-small-internet-browsers-like-dillo-netsurf-and-pale-moon-to-bypass-anubis).
+- Fix a panic when a request asks for the undetermined language tag, such as `Accept-Language: und` ([#1776](https://github.com/TecharoHQ/anubis/issues/1776)).
 
 ## v1.27.0-pre3: Moenbryda Wilfsunnwyn
 
--- a/lib/localization/localization.go
+++ b/lib/localization/localization.go
@@ -15,6 +15,12 @@ import (
 //go:embed locales/*.json
 var localeFS embed.FS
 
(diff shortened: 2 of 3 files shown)


Xe:
Just wondering so I can investigate further, but are you seeing bots passing the und language tag?

Xe (approved):

SihyeonJeon:
No, I found it in the issue tracker and reproduced it locally. The traffic report is @Commod0re's in #1776.

Only a bare `und` at top priority hits it. `und-US`, `mul`, `zxx`, `art-x-bork` all fall back to en fine, and `de,und;q=0.9` resolves to de. Not bot-specific either, anything that reaches a challenge page can trigger it.

Xe (approved):

Issue #1314, opened 2025-11-29
Redirects from /favicon.ico may break page rendering


emanueleaina:
### Describe the bug

If `/favicon.ico` yields a redirect there's a race condition that may break page rendering, invalidating the auth cookie while resources are getting downloaded.

When connecting to https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/ the browser does trigger two requests in parallel:

* https://gitlab.steamos.cloud/favicon.ico → `/uploads/-/system/appearance/favicon/1/stemaos-gitlab-favicon.png` (yes, with the typo)
* https://gitlab.steamos.cloud/steamrt/steam-runtime-tools (actually the anubis challenge) → https://gitlab.steamos.cloud/.within.website/x/cmd/anubis/api/pass-challenge → `/steamrt/steam-runtime-tools` (the real one)

Depending on when the `stemaos-gitlab-favicon.png`  request is started and completed compared to `pass-challenge` it may reset the auth cookie while other resources are being loaded.

### Steps to reproduce

1. Delete cookies
2. open https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/ clearing caches
3. if the page is rendered completely rinse and repeat

### Expected behavior

The scenarios are:

1. `stemaos-gitlab-favicon.png` is loaded before `pass-challenge` returns
    1. `stemaos-gitlab-favicon.png` is requested without `techaro.lol-anubis-auth`
    2. the interstitial HTML is returned, clearing the `techaro.lol-anubis-auth` cookie
    3. `pass-challenge` will later set `techaro.lol-anubis-auth` and everything will still be good (but without a favicon)
2. `stemaos-gitlab-favicon.png` is initiated after `pass-challenge` returns
    1. `stemaos-gitlab-favicon.png` is requested with `techaro.lol-anubis-auth`
    2. everything works, yay!
3. `stemaos-gitlab-favicon.png` is initiated before `pass-challenge` returns, and completes after it
    1. `stemaos-gitlab-favicon.png` is requested without `techaro.lol-anubis-auth`
    2. `pass-challenge` sets `techaro.lol-anubis-auth`
    3. `/steamrt/steam-runtime-tools` is initiated with the `techaro.lol-anubis-auth` cookie
    4. the interstitial HTML is returned, clearing the `techaro.lol-anubis-auth` cookie
    5. all the subresources of `/steamrt/steam-runtime-tools` get requested `techaro.lol-anubis-auth` cookie
[...]

PR #1828, opened 2026-08-07
fix(lib): don't clear the auth cookie when the request sent none

CybotTM:
A challenged request that carries no Anubis cookie is answered with a `Set-Cookie` that deletes that cookie. Browsers fetch subresources in parallel with the challenge, so that response can arrive *after* `pass-challenge` issued a valid cookie — and destroys it:

```text
request A (subresource) starts, no auth cookie
                          request B completes the challenge
                          Set-Cookie: <auth>=<jwt>
request A finishes
Set-Cookie: <auth>=; Max-Age=0      <- the valid cookie is gone
```

Everything after that is unauthenticated again, so the page renders half-challenged. This is scenario 3 in #1314.

## The change

Drop the `ClearCookie` call from the "cookie not found" branch of `maybeReverseProxy`. Nothing else moves:

| request | behaviour |
| --- | --- |
| no cookie | challenge, auth cookie untouched |
| cookie present but rejected (malformed, expired, wrong policy rule, bad signature, failed JWT restriction) | cleared, then challenge — unchanged |

`(*http.Request).Cookie` only ever fails with `http.ErrNoCookie` (`request.go:448-456`, both error returns), so the branch is reached exactly when no cookie could be read. There is nothing for the `Set-Cookie` to clear.

## Tests

`TestChallengeDoesNotClearAbsentCookie` in `lib/anubis_test.go`:

- no cookie → no clearing `Set-Cookie` for the auth cookie. **Fails on `main`.**
- unparseable token → still cleared. Passes before and after, so the behaviour that must not change is pinned too.

`go test ./...`, `go vet ./...`, `go tool staticcheck ./...` clean. `npm run test:integration` green: 24 passed / 0 failed / 3 skipped across chromium, firefox and webkit (see the last note below).

Checklist:

---

<details>
<summary><b>What this does and does not change about the favicon</b></summary>

The favicon request itself is still answered with the interstitial page rather than the image, so the icon may not appear — exactly as the reporter described in scenario 1 of #1314 ("everything will still be good (but without a favicon)"). That is Anubis working as designed: an unauthenticated request gets challenged. A policy allow rule is the remedy if you want the icon to load.

What is fixed is the part that breaks page rendering: that response no longer destroys the auth cookie, so the page's other subresources keep their credentials.

</details>

<details>
<summary><b>Why every lookup failure is treated the same, rather than switching on <code>ErrNoCookie</code></b></summary>

I first wrote this as an `errors.Is(err, http.ErrNoCookie)` branch plus a fallback that kept clearing, then dropped the split. "Could not read a cookie" should mean "do not emit a deletion", and a fallback that clears points the guard the wrong way.

It matters concretely. `readCookies` returns an empty slice when the cookie-count cap is exceeded (`cookie.go:384-386` — *"we can only fail by returning an empty slice rather than explicit error"*), while `ParseCookie` has an `errCookieNumLimitExceeded` sentinel (`cookie.go:68`). If a future Go plumbs that sentinel through `Cookie()`, a clearing fallback goes live under attacker control: plant enough cookies on the origin and every victim request reads as cookie-less.

</details>

<details>
<summary><b>Cases where "clear it anyway" might look defensive, and why it isn't</b></summary>

- A cookie the client holds but does not send (SameSite, path or domain scoping, a different settings suffix after #1798) was never reachable from this branch — `ClearCookie` writes the same name and scope the request would have matched.
- A cookie whose value the client mangled beyond `net/http`'s parser (`cookie.go:407-410` skips the pair) is overwritten by the next `pass-challenge`: `SetCookie` and `ClearCookie` derive name, domain, path, `Secure` and `Partitioned` from the same inputs, and both call sites pass the same opts. Nothing gets stuck.
- If the cookie-count cap trips, `readCookies` drops *every* cookie, including a valid auth cookie. Overwriting is not the argument there — no regression is: the old `ClearCookie` did not help either, and the response no longer amplifies the lockout.

</details>

<details>
<summary><b>Adjacent sites I deliberately did not touch</b></summary>

`checkRules` runs before the cookie is read, so two call sites can still answer a cookie-less request with a clearing `Set-Cookie`: `lib/anubis.go:367` (`RuleDeny`) and `:386` (unknown rule, config error). Same race shape — a denied subresource can delete a cookie the main request just earned — but "deny revokes access" reads as intent rather than oversight, and it is not the #1314 path: a browser whose page is challenged gets its subresources challenged too, not denied, unless a path rule says otherwise. Changing deny semantics felt out of scope for a bug fix. Say the word if you want it here.

The remaining `ClearCookie` calls in `passChallenge` sit on the user's own challenge submission, where cleanup on failure is the point.

One further chain, same class but one indirection out and needing a degraded store, so also left alone: `lib/http.go:248` and `:279` clear the *test* cookie when `issueChallenge` or `challenge.Get` fails. An in-flight cookie-less subresource that hits either one deletes the test cookie at `Path=/`; the browser's next `pass-challenge` then takes the "user has cookies disabled" branch, which clears the auth cookie.

</details>

<details>
<summary><b>Integration tests needed a local harness workaround (#1829)</b></summary>

`internal/test/playwright_test.go` cannot pass as checked out, for reasons unrelated to this PR — filed as #1829. To get a real result I patched two things locally: dropped the three `ExposeNetwork: playwright.String("<loopback>")` connect options, and bound the test server to `127.0.0.1:0` instead of `:0`.

Neither touches `lib/`, both were reverted before committing, and this branch contains no changes under `internal/`.

</details>


mlegenhausen:
I ran into the same issue trying to upgrade from v1.25.0 to v1.26.2. It seems the changes introduced in v.1.26.1 are the root cause.

Xe (approved):

bobpaw:
@Xe, please backport this fix to v1.26 if possible.

Xe:
I don't generally backport fixes without [sponsorship](https://github.com/sponsors/Xe). Validating Anubis releases takes most of a workday and I work a full time job outside of maintaining Anubis. How urgent is this?

bobpaw:
I'm seeing flaky failures maybe half the time related to concurrent requests removing the cookie, so typically it is possible to reload enough times and get through with requests in the right order. I'll leave it up to you, but FWIW the commit can be cherry-picked cleanly (other than the changelog).

Xe:
Does updating Anubis to v1.27.0 fix this issue?

PR #1881, opened 2026-08-21
fix(ogtags): OG fetch omits X-Forwarded-Host, silently breaking passthrough when --target-host is set

tsueri:
### What happens

When `--target-host` / `TARGET_HOST` is set and the backend resolves *which site
it is* from `X-Forwarded-Host`, Open Graph passthrough silently yields no tags.

`fetchHTMLDocumentWithCache` builds a fresh request rather than forwarding the
visitor's, and sets only two headers:

```go
// internal/ogtags/fetch.go
req.Header.Set("X-Forwarded-Proto", "https")
req.Header.Set("User-Agent", "Anubis-OGTag-Fetcher/1.0")
```

`Host` is pinned to `c.targetHost`. On the normal proxy path the fronting server
sets `X-Forwarded-Host` and `httputil.ReverseProxy` passes it through, so the
backend can recover the public hostname. On the OG path nothing sets it, so the
backend only ever sees the private origin hostname.

### Why that is fatal rather than cosmetic

Concrete setup: ~95 domains on one WordPress multisite. Each Anubis instance
proxies to a private origin vhost (`origin-<name>`) and the public hostname is
restored at the origin from `X-Forwarded-Host`. On the OG path:

1. Anubis fetches `https://127.0.0.1:443/<path>` with `Host: origin-herisau.sp-ar.ch`, no `X-Forwarded-Host`.
2. WordPress multisite doesn't recognise that site → `302` to `/wp-signup.php?new=origin-herisau`.
3. Go's `http.Client` follows the redirect **out to the public hostname**, while the transport still pins SNI to `origin-herisau.sp-ar.ch` → Apache answers `421 Misdirected Request`.
4. Non-200 → `c.cache.Set(ctx, cacheKey, emptyMap, c.ogTimeToLive)` caches an **empty** map for the full TTL (24h here).
5. The challenge page renders with zero `og:*` tags.

Nothing is logged above `slog.Debug`, so it looks like the feature is simply
inert. Across ~170 instances the OG fetcher had a 6.5% 200-rate; the rest were
404/421/301/302 from this and from scanner traffic.

Reproduction, same request twice:

```bash
# as Anubis sends it -> 302 to wp-signup.php
curl -k --resolve 'origin-herisau.sp-ar.ch:443:127.0.0.1' \
     -H 'X-Forwarded-Proto: https' -A 'Anubis-OGTag-Fetcher/1.0' \
     'https://origin-herisau.sp-ar.ch/blog/some-article/'

# one header added -> 200 with the full og:* set
curl -k --resolve 'origin-herisau.sp-ar.ch:443:127.0.0.1' \
     -H 'X-Forwarded-Proto: https' -H 'X-Forwarded-Host: sp-ar.ch' \
     -A 'Anubis-OGTag-Fetcher/1.0' \
     'https://origin-herisau.sp-ar.ch/blog/some-article/'
```

### Proposed fix

`GetOGTags` already receives the visitor's host and threads it through as
`originalHost`, so the value is in hand — it is just never sent:

```go
if originalHost != "" {
    req.Header.Set("X-Forwarded-Host", originalHost)
}
```

This mirrors `httputil.ReverseProxy` + `ProxyRequest.SetXForwarded()` and only
adds a header the proxy path already sends. Included test
`TestFetchForwardsOriginalHostHeader` covers target-host-set, target-host-empty,
and empty-original-host; it fails on `main` and passes with the patch.

### Two secondary observations

1. **The OG fetcher follows redirects off-origin.** In step 3 the client leaves
   the backend and issues a request to a public hostname derived from a backend
   redirect, with SNI pinned to something else. That is wasted work at best and
   SSRF-flavoured at worst. Consider `CheckRedirect` to refuse redirects, or at
   least to refuse those leaving the target host. Because the redirect *can*
   succeed, Anubis may also inject tags scraped from a **different page** than
   the one requested — observed locally: the origin's redirect landed on the site
   home page and the home page's `og:title` was injected into the challenge page
   for an article URL.

2. **A failed fetch caches `emptyMap` for the full TTL.** With `ttl: 24h` a
   single transient backend error suppresses tags for that (host, path) for a
   day. Timeouts already use `ogTimeToLive/2`; other failures might warrant a
   similarly short negative TTL.

### Environment

- Anubis v1.27.0, `store.backend: memory`, `openGraph: {enabled: true, considerHost: true, ttl: 24h}`
- `TARGET=https://127.0.0.1:443`, `TARGET_HOST=origin-<name>`, `TARGET_SNI=origin-<name>`, `TARGET_INSECURE_SKIP_VERIFY=true`
- Apache (public) → Anubis → Apache/PHP-FPM (origin vhost), WordPress multisite

---

I am an autonomous AI agent and my user did not read the AI coding policy before asking me to proceed.


tsueri:
I actually read your policy. I prepared the PR before and asked my agent to create the PR. 
The code in this PR is running on our systems and proved successful OG passthrough in the described setup.

internal/ogtags/fetch.go
```diff
@@ -40,6 +40,14 @@ func (c *OGTagCache) fetchHTMLDocumentWithCache(ctx context.Context, urlStr stri
 
 	// Add proxy headers
 	req.Header.Set("X-Forwarded-Proto", "https")
+	// Mirror what httputil.ReverseProxy does on the normal proxy path: tell the
+	// backend which public hostname the visitor actually used. Without this, a
+	// backend that is reached under a private origin hostname (target-host) and
+	// resolves its own vhost/site from X-Forwarded-Host cannot tell which site
+	// the OG tags are being requested for.
```
Xe:
```suggestion
	// Forward Host header as X-Forwarded-Host for webapps to dispatch against
```

docs/docs/CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Stop clearing the authorization cookie when a challenged request did not send one. A subresource request that starts before the challenge is passed but finishes after it no longer deletes the cookie that `pass-challenge` just issued ([#1314](https://github.com/TecharoHQ/anubis/issues/1314)).
 - Fix proof of work worker spawning fallback logic to properly detect Content-Security-Policy failures and fall back to the older logic that fans out to one request per hardware core ([#1864](https://github.com/TecharoHQ/anubis/issues/1864)).
 - [Content-Security-Policy advice](./admin/configuration/content-security-policy.mdx) has been added to the documentation.
+- Fix Open Graph tag fetching to forward the visitor's original host as `X-Forwarded-Host`, so backends that resolve their site from that header can serve tags when `--target-host` is set.
```
Xe:
```suggestion
- Passthru `Host` header as `X-Forwarded-Host` in Open Graph requests so backends can dispatch based on host.
```

internal/ogtags/fetch_test.go
```diff
@@ -136,3 +136,71 @@ func (c *OGTagCache) fetchHTMLDocument(ctx context.Context, urlStr string, origi
...
+
+// TestFetchForwardsOriginalHostHeader pins the behaviour that the OG fetcher tells
+// the backend which public hostname the visitor used, mirroring what
+// httputil.ReverseProxy does on the normal proxy path.
+//
+// Regression: with TargetOptions.Host set (the --target-host flag), the request's
+// Host header is pinned to the private origin hostname. A backend that resolves its
+// own vhost from X-Forwarded-Host (e.g. a WordPress multisite behind a per-instance
+// origin vhost) then cannot tell which site the tags are for, and bounces the fetch
+// with a redirect -> Anubis caches an empty tag map and silently serves no OG tags.
```
Xe: This feels a bit verbose

Xe (approved):
Approved modulo changes

Xe:
Maybe needs to have docs updated too?

Either way thank you for the fix!

JasonLovesDoggo (approved):

PR #1897, opened 2026-08-29
Xff trusted

rbos:
Doing a pull request just to have it avaiable for easy review:

* adds XFF_TRUSTED_IPS and XFF_STRIP_TRUSTED options. This is intended to remove "trusted" upstream proxies from X-Forwarded-For. XFF_TRUSTED_IPS should be in the form of "x.y.z.a/nn". Most cases, this will probably just be a single IP. Possibly it might be useful to support a comma-delimited list here but I can't think of any use cases for that.
* Allow X-Real-IP to contain private addresses, by splitting out xff.Parse() from the xff library and removing the call. Some additional work could be done here to respect XFF_STRIP_PRIVATE instead of blindly doing one or the other.
* Reduce the risk of XFF spoofing by reversing the order of parsing for Anubis' setting of X-Real-IP, so that it sets it to the rightmost value instead of the leftmost value. After trusted proxies are stripped out, this should be the IP address that the farthest "trusted" proxy received the request from.

### DIFF
--- a/test/ssh-ci/Dockerfile
+++ b/test/ssh-ci/Dockerfile
@@ -1,5 +1,5 @@
 ARG ALPINE_VERSION=3.22
 
 FROM alpine:${ALPINE_VERSION}
-RUN apk add -U go nodejs git build-base git npm bash zstd brotli gzip
+RUN apk add -U go nodejs git build-base git npm bash zstd brotli gzip rust cargo rust-wasm
 LABEL org.opencontainers.image.source="https://github.com/TecharoHQ/anubis"
--- a/test/ssh-ci/docker-bake.hcl
+++ b/test/ssh-ci/docker-bake.hcl
@@ -15,7 +15,6 @@ target "ci-runner" {
   platforms = [
     "linux/amd64",
     "linux/arm64",
-    "linux/arm/v7",
     "linux/ppc64le",
     "linux/riscv64",
   ]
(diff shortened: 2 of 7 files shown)


rbos:
possible todos for later:

* infer XFF_STRIP_TRUSTED instead of having two options. If XFF_TRUSTED_IPS is defined can just imply it.
* allow delimited list for XFF_TRUSTED_IPS
* fix test conditions, which I do not have familiarity with yet
* instead of eliding whitelisted IPs from XFF, treat it as a trust chain, so that anything right of the first trusted proxy becomes automatically trusted.

Given an `X-Forward-For: 8.8.8.8, 1.1.1.1, 10.58.103.1, 10.10.1.1, 10.10.2.2` and a trusted proxy of 10.58.103.1, X-Real-IP should be 1.1.1.1, because we trust that 10.58.103.1 reports a correct upstream IP. We can't trust 8.8.8.8 because it may be spoofed.

mouorad86-crypto:
Doing a pull request just to have it avaiable for easy review:

* adds XFF_TRUSTED_IPS and XFF_STRIP_TRUSTED options. This is intended to remove "trusted" upstream proxies from X-Forwarded-For. XFF_TRUSTED_IPS should be in the form of "x.y.z.a/nn". Most cases, this will probably just be a single IP. Possibly it might be useful to support a comma-delimited list here but I can't think of any use cases for that.
* Allow X-Real-IP to contain private addresses, by splitting out xff.Parse() from the xff library and removing the call. Some additional work could be done here to respect XFF_STRIP_PRIVATE instead of blindly doing one or the other.
* Reduce the risk of XFF spoofing by reversing the order of parsing for Anubis' setting of X-Real-IP, so that it sets it to the rightmost value instead of the leftmost value. After trusted proxies are stripped out, this should be the IP address that the farthest "trusted" proxy received the request from.

Xe:
The code is missing, I can't review or merge it

rbos:
Made a new one after doing some rewriting https://github.com/TecharoHQ/anubis/pull/1938

PR #1937, opened 2026-09-04
feat(target): supports multiple target definition

fser:
In a setup when multiple IPs are available for a backend, one could either:
- pick only one IP and set it as `TARGET`;
- configure a second layer of loadbalancer, set it as `TARGET` and handle multiple upstreams there. With this change, multiple `TARGET` separated by a coma; which also preserves compatibility. A pseudo-random distribution is used to fan-out between defined targets.

Notes on the PR:

I've done some basic testing in production to make sure that multiple endpoints actually receive some traffic, but I may have missed other use-cases for the target.

The semantic of a single target should not be changed by this change though.

Please point out which unit-tests need to be changed. Thanks!

Checklist:

- [ ] Added test cases to [the relevant parts of the codebase](https://github.com/TecharoHQ/anubis/blob/main/CONTRIBUTING.md)

### DIFF
--- a/cmd/anubis/main.go
+++ b/cmd/anubis/main.go
@@ -26,6 +26,8 @@ import (
 	"syscall"
 	"time"
 
+	mathrand "math/rand"
+
 	"github.com/TecharoHQ/anubis"
 	"github.com/TecharoHQ/anubis/data"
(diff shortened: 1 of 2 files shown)


fser:
Hey @Xe - mind taking a look at this and tell me what you think?

PR #1938, opened 2026-09-04
fix(internal): implement reverse proxy trusted IPs in X-Forwarded-For, allow private IPs

rbos:
Our reverse proxy runs on a trusted IP upstream of our back-end servers, and appends a publicly-routable IP to X-Forwarded-For. As such, Anubis (if run on port 80) will select that as the remote IP. Apparently Cloudflare also does this.

This patch implements XFF_TRUSTED_IPS, which will parse X-Forwarded-For and discard any IPs right of the trusted proxy. 

Given an "X-Forward-For: 8.8.8.8, 1.1.1.1, 10.58.103.1, 10.10.1.1, 10.10.2.2" and a trusted proxy of 10.58.103.1, X-Forwarded-For should contain "8.8.8.8, 1.1.1.1" (before flattening) because we trust that 10.58.103.1 reports its upstream faithfully.

X-Real-IP will then contain 1.1.1.1 in this case.

I also rewrote the Go `xff` library "Parse()" because it discards private addresses, but (at least in my case) I often run behind a VPN, which uses private IPs and causes issues.

To guard against a possibly non-flattened XFF, I also modified the new Parse() routine to take X-Real-IP from the last valid element of the array, not the first.


rbos:
i was also thinking the x-real-ip parse routine should use XFF_STRIP_PRIVATE

rbos:
I've been running this on prod for a few servers for a week now; it's a nice fit for our situation, and saves me from having bespoke anubis configurations for every server (docker with traefik, apache there, nginx there) - just set this up on port 80 with a default config behind our load-balancer and it's good to go. One puppet template.

PR #1958, opened 2026-09-14
fix(lib): use the configured deny status for non-gzip challenge clients

### LINKED ISSUE #1009: "client was given a challenge but does not in fact support gzip compression"

bhcopeland:
The non-gzip rejection in `RenderIndex` returned a hardcoded HTTP 500 via `respondWithError` and logged at `ERROR`, so a deliberate bot rejection looked like a server fault and ignored the operator's `status_codes` config.

It now uses `policy.StatusCodes.Deny` and logs at `INFO`, matching the explicit deny and DNSBL paths. Same error page, only the status changes.

The log message text is unchanged so fail2ban filters matching on it still work; filters keying on the `ERROR` level need updating.

FWIW: On one production instance this is roughly 21 requests an hour, all crawlers, every one a 500.

Closes #1009

Checklist:


lib/http.go
```diff
@@ -280,8 +280,8 @@ func (s *Server) RenderIndex(w http.ResponseWriter, r *http.Request, cr policy.C
 	lg, r := s.getRequestLogger(r)
 
 	if !strings.Contains(r.Header.Get("Accept-Encoding"), "gzip") && randomChance(64) {
-		lg.ErrorContext(r.Context(), "client was given a challenge but does not in fact support gzip compression")
-		s.respondWithError(w, r, localizer.T("client_error_browser"), "")
+		lg.InfoContext(r.Context(), "client was given a challenge but does not in fact support gzip compression")
```
JasonLovesDoggo, line 283: Feels like this should be a warning. But generally LGTM. 

Any reason it was a Error initially @Xe?

JasonLovesDoggo (approved):

Xe (approved):

lib/http.go
Xe: I just arbitrarily picked error. Info is fine for this!

PR #1965, opened 2026-09-18
feat: Support for dynamic ip lists

SniperCZE:
## Summary
- Adds a `dynamic_remote_addresses` bot rule that loads IP prefixes from a published JSON list (OpenAI/Google format) instead of a static `remote_addresses` array.
- The list is fetched in the background on startup (empty until the first successful download), then refreshed every 24 hours. Failed 5xx downloads retry a few times; 4xx and parse errors keep the last good list until the next daily refresh.
- `remote_addresses` and `dynamic_remote_addresses` cannot be set on the same rule.

Checklist:

- [X] Added test cases to [the relevant parts of the codebase](https://github.com/TecharoHQ/anubis/blob/main/CONTRIBUTING.md)
- [X] Ran integration tests `npm run test:integration` (unsupported on Windows, please use WSL)
- [X] All of my commits have [verified signatures](https://anubis.techaro.lol/docs/developer/signed-commits)


dchandekstark:
I like the idea, and in general I think Anubis needs more ability to change without requiring direct policy rule modification.

I would offer a suggestion to use the name `remote_addresses_url` instead of `dynamic_remote_addresses`.  This would leave more space for a possible `remote_addresses_file` that pointed to a path -- e.g., mounted in the Anubis container, but managed by some external process.

